PR #10, opened 2019-05-28
WIP: Bindings to useEffect Hook

benschinn:
Addressing issue https://github.com/jchavarri/rroo/issues/6.

May-28-2019 21-45-28


lib/React.js
```diff
@@ -39,3 +39,12 @@ var useReducer = function(reducer, initialState) {
   var tuple = React.useReducer(jsReducer, initialState);
   return caml_js_to_array(tuple);
 };
+
+// Provides: useEffect
+// Requires: React, caml_js_wrap_callback, caml_js_from_array
+var useEffect = function(effect) {
+  var callback = caml_js_wrap_callback(
+    effect === 0 ? null : caml_js_from_array(effect)[0]
```
jchavarri: This is kind of tricky. What this code is doing is assuming that the effect callback itself is optional, but what we want is to convert the optional params that exist "inside" the callback.

In order to do so, one can create an inline function here, that is the one that is passed to `React.useEffect`. And then inside that callback, convert the required parameters.

I am just working on a similar example for `forwardRef`:

https://github.com/jchavarri/rroo/blob/6f6ac6b39bcd2d884a81b8e4f39de3d2869c6742/lib/React.js#L61-L67

Hope this helps!

jchavarri: Feel free to copy the `optFromJs` from that branch, or inline it as it was before, we can consolidate later on 👍

benschinn: Yay! I got the js implementation working correctly. Thank you for the pointer! Btw on the code snippet you shared above, is the `jsCallback` function supposed to take `ref` as a param instead of `jsRef`? It appears variable `ref` is unused.

lib/React.re
```diff
@@ -113,6 +111,41 @@ external useReducer:
   (('state, 'action) => 'state, 'state) => ('state, 'action => unit) =
   "useReducer";
 
+external useEffect: 
+  (unit => option(unit => unit)) => unit = 
+  "useEffect";
+
+// external useEffect0:
+//   (unit => option(unit => unit), Js.array_empty) => unit =
```
benschinn: I am not sure how to express an empty array for `useEffect0`. I did find [Js.array_empty](https://ocsigen.org/js_of_ocaml/3.1.0/api/Js) on the jsoo docs but how do I use it here? I tried `Js_of_ocaml.Js.array_empty` but to no avail.

lib/React.js
jchavarri: Great!

> It appears variable `ref` is unused.

Hm... you're right, good catch! I will open an issue to revisit `forwardRef`.

As an aside, I am thinking if we could leverage TypeScript or eslint to keep track of these very simple errors. I'll create a new issue for that too 👍

lib/React.re
jchavarri: With BuckleScript bindings, which are kind of "inlined", we have to find a way to define constant values in the signature.

With the `external` approach of rroo that splits type definitions with actual implementation, we don't need to 🙂 We can just skip the dependencies param from the signature, and then in the `.js` implementation pass the empty array. ✨

jchavarri: For an example of something similar, check how the `createFragment` function doesn't take any constant for the element type, just the array of children:
https://github.com/jchavarri/rroo/blob/7f5e9626742c9b2d8c602fb92c26544c7174d36e/lib/ReactDOM.re#L46

but later on, in the JS implementation, we read `React.Fragment` to pass it as the first param of `React.createElement`:

https://github.com/jchavarri/rroo/blob/7f5e9626742c9b2d8c602fb92c26544c7174d36e/lib/ReactDOM.js#L39-L43

benschinn: Ah I see. Would that mean the `useEffect` and `useEffect` would have the same external declarations? In that case would we not want `useEffect0`? Or do we want to keep it to keep API as close as possible to ReasoReact?

benschinn: Actually I am doing this for `useEffect` and it seems to be working:

`external useEffect0: (unit => option(unit => unit), array(unit)) => unit = "useEffect";`

jchavarri: Thanks a lot for your work on this! 🙌 

> Ah I see. Would that mean the `useEffect` and `useEffect` would have the same external declarations? In that case would we not want `useEffect0`?

They would have the same type signatures but compiled down to different behavior, via the JavaScript binding code. The idea is:

- external `useEffect` -> function `useEffect` in `React.js` file -> calls `React.useEffect`
- external `useEffect0` -> function `useEffect0` in `React.js` file -> calls `React.useEffect` with an empty array

One could think of these two functions as `useEffectAlways` (for `useEffect`) and `useEffectOnce` (for `useEffect0`) if that makes sense. Although I would tend to think it's beneficial to keep the naming close to ReasonReact, for familiarity.

> Or do we want to keep it to keep API as close as possible to ReasonReact?

I think it's interesting to stay as close as ReasonReact as possible, but there is room to explore and tweak the APIs to make them more ergonomic, or the types more OCaml-idiomatic.

> Actually I am doing this for `useEffect` and it seems to be working

In this case, I feel more inclined to define `useEffect0` without the dependency param, because being a constant value, the `array(unit)` type might lead to confusion (what is an array of nothing? nothing 😄 ).

One important piece of context: In ReasonReact case, they chose those types because they had a very hard constraint on zero cost in the resulting bundle size, but in rroo that constraint does not exist 🙂

benschinn:
@jchavarri I tried resolving conflict here, but I get an infinite loading state on the page. I pulled down and merged master and resolved conflict locally, but doesn't seem to be passing still. Do I need to undo my merge? Or are you able to resolve the conflict here?

jchavarri:
hey @benschinn !

I'm not sure what problems you ran into, I was not able to repro, maybe it was some local cache thing? I opened it and it works beautifully! 😍 

I fixed the conflicts and pushed also a small change of "promise-like" delay behavior with [`Lwt`](https://github.com/ocsigen/lwt), which is OCaml's library to handle asynchronicity. I think your work with effects and this Lwt usage shows really well the potential of leveraging the OCaml ecosystem with React in a seamless way.

One thing I noticed is that the "unmount" effects that are returned in the `option` don't seem to run 🤔 I guess it's due to some conversion that is remaining.  I think though that this PR already includes a lot of upsides on its own, so I'll open a new issue to track that separately.

Thanks again for your work @benschinn ! Very exciting to have support for effects in rroo now 🚀

benschinn:
@jchavarri Thanks for guiding me with the first PR! 😃

I have been thinking about what React implementation on jsoo might look like. You've done a beautiful work so far. I am looking forward to contributing more!

P.S. - I will do another PR for removing the `array(unit)` from the `useEffect0` signature and add a `useEffect0` function to the js binding.

Issue #14, opened 2019-05-29
Add linting on js files


jchavarri:
There are some basic mistakes that can happen all the time in `.js` files. Adding eslint (or typescript?) would help mitigate that.

jchavarri:
We could also use BuckleScript (kidding / not kidding ) 😄

benschinn:
lol, I kinda like the idea of using BuckleScript. that would take care of both linting and type.

jchavarri:
Yeah 😄 I have been thinking more, and if extra tooling is going to be added, we could just as well use jsoo [`Js` module](http://ocsigen.org/js_of_ocaml/3.1.0/api/Js), and ppx (`obj##.property`, `obj##method`), which has quite good support to deal with JS data types (actually BuckleScript was inspired from it in many ways).

As an example, this:

https://github.com/jchavarri/rroo/blob/4aac16dac3c6fed395805b33c4e0ae6035a27ce6/lib/InteropUtils.js#L11-L13

would become this:

```reason
Js.Opt.to_option
```

😂 

Maybe a small experiment with `useState` for example would throw more light, but it seems like the simplest path forward. If that was the case the `external` go away and are replaced with regular `let` where the converted logic for each function would go.

benschinn:
I wish I had more experience with jsoo to give constructive input. Although it's very fascinating to learn where bs drew some inspiration. 😅

So if I understand correctly, if we decided to just use the jsoo, `lib/React.js` would be replaced with one jsoo ffi implementation? Any no need for `external`s ?

jchavarri:
@benschinn Exactly! No need for `.js` files anymore.

PR #18, opened 2019-06-01
Experiment: Use Js_of_ocaml.Js to write bindings

jchavarri:
As part of the discussion in #14 one idea was to write the bindings using jsoo `Js` module.

This PR explores the idea with a few functions. There is though a lot of unsafety and escape hatches, plus it requires knowing and understanding how all these `Js_of_ocaml.Js` functions and types work.

A couple of examples:
- Calling a method in an object has a type signature `'a => Js.meth('b)`. 😕 
- There are a lot of calls to `Js.Unsafe` to convert from Js values. For example when transforming the arrays returned from `useState` or `useReducer` to tuples. I don't think there is a safe way to do this conversion, or at least I couldn't find anything.

I am not sure this is the right path to follow. It seems to impose a much higher barrier of entry to potential contributors and the theoretical gains in safety are mostly gone due to the unavoidable unsafety in many cases.

@benschinn I wonder what your thoughts are? 🙂

### DIFF
--- a/lib/React.c
+++ b/lib/React.c
@@ -1,15 +1,5 @@
 #include <stdlib.h>
 #include <stdio.h>
-void nullElement () {
-  fprintf(stderr, "Unimplemented React function `nullElement`!\n");
-  exit(1);
-}
-
-void stringElement () {
-  fprintf(stderr, "Unimplemented React function `stringElement`!\n");
-  exit(1);
-}
-
 void createElement () {
   fprintf(stderr, "Unimplemented React function `createElement`!\n");
   exit(1);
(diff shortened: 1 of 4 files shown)


benschinn:
Thanks for tagging me here, I was very interested in what using just jsoo would look like. 😃

I really like that this reduces or eliminates `.js` and `.c` code, and you only need to reference the `lib/React.re` and interface file. 

I do agree that this approach could be a barrier for potential contributors. For example the [`converKeyToJs`](https://github.com/jchavarri/rroo/pull/18/files#diff-b7abd0a935cd0129b16d3f73be220b0bR24) is much less idiomatic and doesn't appear very typesafe. 

I will have to mull over  a bit more what the best approach might be.

jchavarri:
@benschinn Thanks for your feedback! I really appreciated it.

Based on this, I am inclined to explore next how a non-jsoo solution would look like. It seems that a lightweight eslint config (that we can tweak as needed) could take us a long way 🤔

jchavarri:
Closing this as eslint has been set up already by @benschinn in #20 :)

Issue #4, opened 2019-05-27
Add CI


jchavarri:
https://github.com/esy-ocaml/hello-reason is a good example on how to set Azure pipelines.

Would be nice to test at least the ppx.

PR #20, opened 2019-06-04
Eslint Config

### LINKED ISSUE #14: Add linting on js files

benschinn:
Fixes #14. In reference to conversation with @jchavarri in PR https://github.com/jchavarri/rroo/pull/18:

A lightweight eslint config! 

In `lib` directory run:
>  `$ yarn lint`

P.S.- sorry for the misspelled branch name meant to name it `js-lint`


lib/package-lock.json
```diff
@@ -0,0 +1,1015 @@
+{
```
jchavarri: I would prefer to run `yarn` instead and keep one single lock file, as there's a `yarn.lock` already. In the future, once esy supports PnP, we might be able to do everything with esy.

lib/.eslintrc
```diff
@@ -0,0 +1,13 @@
+{
+  "parserOptions": {
+      "ecmaVersion": 5
+  },
+  "extends": "eslint:recommended",
+  "rules": {
+    "no-undef": "warn",
```
jchavarri: What would be the best way to deal with js_of_ocaml functions like `caml_js_to_array` that are available globally? Should we add them to some global file so eslint ignores them?

lib/.eslintrc
```diff
@@ -0,0 +1,13 @@
...
+    "no-unused-vars": "warn"
```
jchavarri: How to avoid warnings from bindings functions like `Ref_current` that are exported from a `.js` file?

lib/package.json
```diff
@@ -1,6 +1,12 @@
 {
+  "scripts": {
+    "lint": "eslint React.js"
```
jchavarri: Should we include `ReactDOM` here too? (actually, the preferred script would check _any_ `.js` file that is in `lib` and it's not in `node_modules`).

benschinn: good point! I replaced it with `eslint *.js`, now linting all .js files in lib directory:
https://github.com/jchavarri/rroo/pull/20/files#diff-b6397f6861def56ce657a2dde39e2cbbR3

lib/.eslintrc
benschinn: Added global variables:
https://github.com/jchavarri/rroo/pull/20/files#diff-b0d0bf091f28af6d677bffef8898431bR10

I turned the `"no-undef"` rule to `"error"`, so when we run eslint next time we add new global variables we will know to add it to the `.eslintrc`.

benschinn:
@jchavarri hey thanks for the review! I pushed up some changes that addresses most things I think. 
Screen Shot 2019-06-04 at 7 57 26 AM

Now getting just one warning from lint. And it might actually be a bug.

It looks like `value` argument is unused:
https://github.com/jchavarri/rroo/blob/master/lib/React.js#L71

lib/package-lock.json
jchavarri: Yeah, `esy.lock` needs to be commited, but I think the current setup you have with yarn.lock inside `lib` is perfect 👌  until we revisit the packaging / dependencies handling with `esy` in the future.

benschinn:
@jchavarri should we use something like [husky](https://github.com/typicode/husky) for linting on  a pre commit hook?

lib/.eslintrc
```diff
@@ -0,0 +1,28 @@
...
+  "parserOptions": {
+      "ecmaVersion": 5
+  },
+  "extends": "eslint:recommended",
+  "rules": {
+    "no-undef": "error",
+    "no-unused-vars": ["warn", { "vars": "local" }]
+  },
+  "globals": {
+    "joo_global_objec": "readonly",
```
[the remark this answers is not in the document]
benschinn: removed it!
https://github.com/jchavarri/rroo/pull/20/commits/ba3b4e869688cb1ddb24e486b8fe0e850924240e

benschinn:
oops. I misunderstood a comment above and committed the `esy.lock` directory. should I remove it?

jchavarri:
> should we use something like husky for linting on a pre commit hook?

@benschinn Husky is cool! Although, I would only add it if we run into linting issues pretty often, for now I'd feel inclined to keep things leaner, especially considering we'll have soon CI integration when #4 is done.

> I misunderstood your comment above and committed the esy.lock directory. should I remove it?

Ah! yes please 🙏 There's no rush though, I can do it later otherwise :)

benschinn:
> Husky is cool! Although, I would only add it if we run into linting issues pretty often, for now I'd feel inclined to keep things leaner, especially considering we'll have soon CI integration when #4 is done.

Sounds good! 

> Ah! yes please 🙏 There's no rush though, I can do it later otherwise :)

Done! https://github.com/jchavarri/rroo/pull/20/commits/e41bcf88a5c19de1855ab987015a3f125f6957bc

jchavarri:
Thanks @benschinn ! ✨

PR #23, opened 2019-06-22
Upgrade to OCaml 4.07

jchavarri:
Updates the version of the OCaml compiler to 4.07.

Note: There is a missing part before this can be merged. Right now there is a hard dependency on jsoo ppx, which is used as a library:

https://github.com/jchavarri/rroo/blob/47ed88f469f4e853b07dcd6c5a3a6687a035a5c9/ppx/Rroo_jsoo_ppx.re#L1138-L1142

This breaks encapsulation, and forces rroo to follow exactly the same version of the compiler that ` 	js_of_ocaml-ppx` is using.

jsoo updated to 4.07 recently https://github.com/ocsigen/js_of_ocaml/pull/762 so maybe it's just a matter of waiting until opam package is updated.

cc @hhugo :)  are those assumptions correct? Would there be a way to remove that dependency on jsoo ppx without complicating this one a lot?


jchavarri:
Superseded by #25.

PR #31, opened 2019-10-13
fix example cd path

naartjie:
going through the instructions for setting up the example, and thought I would fix the path.


jchavarri:
Thanks! :)

Issue #47, opened 2020-10-14
Compiling without Webpack


tmattio:
Hi @jchavarri!

I've been setting up `jsoo-react` in my project and thought I'd give some feedback on the setup 🙂 

To give you some context, I'm using `ocaml-crunch` to generate OCaml modules for my static content. It simplifies the deployment drastically: with this setup, I can just ship an executable and don't have to worry about storing static files.
To serve the React app, I am embedding the compiled Js in an OCaml module and serve it from one of the endpoints of the server.

Now, all of this is pretty much automated with Dune configurations, however, the React app first needs to be processed by Webpack (or another bundler), because the code generated by Js_of_ocaml imports `react`.

This dependency on Webpack adds a lot of overhead to the setup, and I'm wondering if it would be possible to get rid of the imports in the generated Js code.

The HTML page could potentially get the react library from their CDN:

```
<script crossorigin src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
<script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
```

And the compiled Js could just assume that `React` is available.

I'm not very familiar with Js_of_ocaml, so I have no idea if that's feasible, but if it is, it would have some nice benefits:

- Simplify the setup by removing the dependency on Webpack.
- Improve build times, because we wouldn't need Dune to check the `node_modules`.
- (Maybe?) this would simplify Server Side Rendering if one-day `jsoo-react` uses `Tyxml`

What do you think?

jchavarri:
I think it's a great idea! I wouldn't like to remove Webpack support, because it might be interesting in the future (for things like lazy loading) and I think some folks will appreciate that jsoo-react has some guidelines on how to be used with Webpack.

But otherwise, I love the simplification and build times improvements :) Your proposal made me realize we have no good reason to assume anything about how React gets provided to final bundles, we can just do nothing inside the library and resolve the globals in userland. I created a PR in #48, let me know what you think!

jchavarri:
Fixed in #48.

PR #48, opened 2020-10-14
Allow consumers to decide how React is provided

jchavarri:
Originated by #47.

There is no good reason why the `require()` calls are part of the library dune files. Consumers can decide how React will be provided, as long as it is on the globals that are expected to be read.

### Webpack

Can add `(javascript_files react.js react-dom.js)` to their application `dune` files, and then `require` inside (or `import`, whatever they choose).

### UMD

Can add `<script>` tags before loading the js file generated by jsoo (what this PR shows)

@tmattio Does this go in the right direction? If yes, I think I could just remove the `__LIB__` prefixes (which never convinced me tbh 😄 ) so that there's no need to alias anything.


tmattio:
Thanks a lot for looking into it @jchavarri!

That completely removes the overhead I was mentioning in the issue, so exactly what I thought 🎉 

For reference, here's a dune configuration I tried that uses both the Webpack setup and the Jsoo-only setup:

```
(executable
 (name index)
 (modules (:standard \ index_dev))
 (modes js)
 (libraries jsoo-react.lib my_lib)
 (preprocess
  (pps jsoo-react.ppx)))

(executable
 (name index_dev)
 (modules (:standard \ index))
 (modes js)
 (js_of_ocaml
  (javascript_files react.js react-dom.js))
 (libraries jsoo-react.lib my_lib)
 (preprocess
  (pps jsoo-react.ppx)))
```

jchavarri:
thanks for the examples @tmattio ! I added some instructions to set up the library with or without Webpack in the readme:
https://github.com/jchavarri/jsoo-react/blob/bdecac1d9685afdc705fbd2c2c9d9e49a84cda48/README.md#existing-project

Gonna merge this now, but if you think there are better ways to solve this, please let me know :)

PR #64, opened 2021-09-29
Switch to ppxlib

keremc:
- Switch to ppxlib
- Drop the dependency on ocaml-migrate-parsetree
- Switch other libraries & tools to ppxlib-compatible versions

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -19,13 +19,9 @@
    `React.createFragment([foo])`
  *)
 
-open Migrate_parsetree
-open OCaml_410.Ast
+module Ocaml_location = Location
+open Ppxlib
 open Ast_helper
-open Ast_mapper
-open Asttypes
-open Parsetree
-open Longident
 
 let rec find_opt p = function
   | [] ->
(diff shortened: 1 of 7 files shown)


ppx/ppx.ml
```diff
@@ -231,7 +227,7 @@ let filenameFromLoc (pstr_loc : Location.t) =
   let fileName =
     match pstr_loc.loc_start.pos_fname with
     | "" ->
-        !Location.input_name
+        !Ocaml_location.input_name
```
jchavarri, line 230: Could use `Ppxlib.Expansion_context.Base.input_name ctxt`?

jchavarri (approved):
Thanks a lot for this PR 👍  I just have a small suggestion to stop relying on `Ocaml_location`, but otherwise looks great.

The tests also seem to be passing: https://github.com/reason-in-barcelona/jsoo-react/actions/runs/1377560682 (there's an unrelated issue with mac build).

I'm gonna go ahead and merge it, will fix the location thing afterwards. Thanks again!

PR #66, opened 2021-11-19
Inline props as object

### LINKED ISSUE #43: Remove superfluous JS output from ReactDOM.domProps
The `domProps` "builder" function has a ton of attributes:

https://github.com/jchavarri/jsoo-react/blob/631f9855c727994b12aa928116b9b7068ae15e5c/lib/reactDOM.mli#L65-L75

The resulting JS code generated by jsoo includes zeroes for all of them missing, which is just useless information that increases for every host element rendered (e.g. `<div />`).

There are some solutions / workarounds that I can think of:

1. Reduce the amount of attributes (super short term 😅 ) by wrapping `<div />` with a "composite" (non host) component like `Div`, passing only the props the app will use.

3. "Lift" the object creation up to the ppx, which has already all the information available. So instead of calling `domProps` here:

https://github.com/jchavarri/jsoo-react/blob/631f9855c727994b12aa928116b9b7068ae15e5c/ppx/ppx.ml#L528-L533

we would do something like:

```ocaml
match nonEmptyProps with
| (Labelled "onClick", expr) :: _ -> (* actually, mapping over the `nonEmptyProps` list, not matching it :) *)
    let mapped_expr = mapper.expr mapper expr in
    [%expr
      let empty = Ojs.empty_obj () in
      Ojs.set empty "onClick"
        (Ojs.fun_to_js 1 (fun inner ->
              [%e mapped_expr] (ReactEvent.Mouse.t_of_js inner))) ;
      Obj.magic empty]
```

3. Explore using [tyxml](https://github.com/ocsigen/tyxml), which compiles the attributes down to a list, a representation that doesn't have this issue.

davesnx:
Fixes #43.

This PR changes the interface for **`React.Dom.domProps`**. Before these changes, `domProps` was a function with ~400 optional labelled arguments, after these changes we treat it as an object.

There are a few reasons for this change, but the most important is bundle size. jsoo compiles the optional arguments into zeros generating per each `domProps` call near 400 zeros, jsoo treated labeled arguments as positions into arrays. There are a few issues opened about this, I only found this one though: https://github.com/ocsigen/js_of_ocaml/issues/801

Another of the reasons is about correctness, `props` in React are an object and after those changes, we treat it as an Object as well. Giving the same flexibility as the underneath layer could enable features that aren't possible on reason-react or rescript/react, for example adding missing attributes, allowing destructuring props in bindings, allowing "data"-attributes or type-safety on certain attributes.

### Todos
- [x] Basic functionality with attributes
- [x] What should we do with `ref` and `key`
- [x] Type-safe attributes
- [x] Type-safe events
- [x] Add `suppressContentEditableWarning`
- [x] Add `dangerouslySetInnerHTML`
- [x] Update snapshot tests
- [x] Add a few test cases
- [x] Fix test `baseSuite:refs:forwardRef` (maybe it's related with my doubt? probably no)
- [x] Analyze better the bundle-size
- [x] Re-visit `dangerouslySetInnerHTML` type/format.
- [x] Fix locations for unexpected prop
- [x] Understand why there's nonlabelled props there?
 
### Doubts
I have a doubt about the `ref` prop. Right now the ref is typed with `React.Dom.Ref.t option` but `React.Dom.domRef` doesn't seem to return an optional. But I believe the bindings are not the same as reason-react since they use an anonymous type from the useRef.

```
File "example/src/Code.re", line 14, characters 14-45:
14 |     <code ref={React.Dom.Ref.domRef(codeRef)}> {text |> React.string} </code>
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type React.Dom.domRef
       but an expression was expected of type React.Dom.domRef option
```

### Future

As mentioned before, this change allows us a few niceties on the dom attributes API, from the top of my head:

- -We could add more type-safety on many properties:
    - polyvariants (for example `?preload is now string (* "none", "metadata" or "auto" (and "" as a synonym for "auto") *)` into `'None | 'Metadata | 'Auto)`
    - dates (for example: `max (* should be int or Js.Date.t *)` into Js.Date
    - regex
    - etc
- We could add warnings/errors about deprecated attributes
- Try to implement "data" attributes (if the propName starts with "data", we render the value without any typecheck.

PS: I have the editor configured to format using ocamlformat, I believe we should run it across all files (I separate the formatting into one commit, in order to not affect the review)


ppx/ppx.ml
```diff
@@ -510,30 +566,41 @@ let jsxMapper () =
...
+          let propsWithName =
+            List.filter (fun (name, _) -> getLabel name != "") nonEmptyProps
+          in
+          let makePropField (arg_label, value) =
+            let name = getLabel arg_label in
+            let prop = Html.findByName name in
+            [%expr
+              [%e
+                Exp.constant ~loc (Pconst_string (Html.getHtmlName prop, None))]
+              (* loc here points to the element <div />, we could be more precise and point to the prop *)
```
jchavarri, line 591: Maybe you can read `pexp_loc` from the label expression that is attached to each element in `nonEmptyProps`? 🤔

jchavarri (approved):
@davesnx This is super cool. 😍 

Re: the refs, there is some context about these decisions in https://github.com/reasonml/reason-react/issues/421#issuecomment-498011419. Essentially `forwardRef` is unsafe in ReasonReact so I tried to fix that by introducing a new type for refs for dom elements (`Dom.Ref.t`).

After your changes, I just realized that we could keep the ref as non-optional, and actually it makes sense to do so, because in some cases (like when using `useRef`) we get a `{current: elementOrNull}` JS object, so converting back and forth was not trivial.

I updated the tests as well, and checked that the ref examples were working correctly.

Another thing, I checked the bundle size of the example (built with `profile=prod`) and we went **from 200KB to 117KB** for the file `App.bc.js`. 🎉 This is all application code (not react or prism libs), but I think it's a huge improvement.

> Giving the same flexibility as the underneath layer could enable features that aren't possible on reason-react or rescript/react, for example adding missing attributes, allowing destructuring props in bindings, allowing "data"-attributes or type-safety on certain attributes.

This is SO exciting.

Thanks for working on this! 🤝

jchavarri:
This is a quick analysis of bundle size impact of this PR using webpack-bundle-analyzer.

For each case, I ran:
- `make build-prod` from project root folder
- Then `yarn webpack:production` from `examples` folder (previously adding `new BundleAnalyzerPlugin()` to webpack plugis list.

This branch (https://github.com/reason-in-barcelona/jsoo-react/pull/66/commits/84e7869ba23ec3c7e9be85233cc4da4318a85920):
image

Master branch (https://github.com/reason-in-barcelona/jsoo-react/commit/1ff26837c94142005a61fea5316df0028529132d):
image

ppx/ppx.ml
```diff
@@ -434,7 +490,7 @@ let jsxMapper () =
     let recursivelyTransformedArgsForMake =
       argsForMake
       |> List.map (fun (label, expression) ->
-             (label, mapper.expr mapper expression))
+             (label, mapper.expr mapper expression) )
```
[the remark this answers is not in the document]
davesnx: Right, I ran with ocamlformat version 0.19.0.

ppx/html.ml
```diff
@@ -0,0 +1,584 @@
+type attributeType = String | Int | Float | Bool | Style | Ref | InnerHtml
```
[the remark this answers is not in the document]
davesnx: I moved out of dom.mli since those get exposed on purpose to the user, while the html.ml/i shouldn't be. If makes more sense to keep it inside Dom, we can make a module called... Private__ or somekind to express it better. WDYT?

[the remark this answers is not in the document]
jchavarri: Ah right, I didn't realize the `html.ml` file was in `ppx` folder. 👍 

Maybe just inline this inside this module in `ppx.ml` then? So it's clear that is something internal to the ppx.

About the modlue naming, I would use `module Dom` for consistency with the other one, and with [React.js docs](https://reactjs.org/docs/dom-elements.html). https://reactjs.org/docs/dom-elements.html, as HTML has different attributes than the props passed to these elements.

But these are all nits :)

ppx/ppx.ml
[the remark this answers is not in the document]
jchavarri: fwiw we run `make format` now on ci to make sure code is always well formatted.

ppx/ppx.ml
```diff
@@ -510,30 +566,41 @@ let jsxMapper () =
...
-          let propsCall =
-            Exp.apply ~loc
-              (Exp.ident ~loc
-                 {loc; txt= Ldot (Ldot (Lident "React", "Dom"), "domProps")})
-              ( nonEmptyProps
-              |> List.map (fun (label, expression) ->
-                     (label, mapper.expr mapper expression)) )
+          (* Filtering out the props that don't have a label, not sure how's possible *)
+          let propsWithName =
+            List.filter (fun (name, _) -> getLabel name != "") nonEmptyProps
```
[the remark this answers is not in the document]
davesnx: Probably the last prop as unit? I'm not sure actually. Will look into it and not merge this until it's "discovered/fixed".

ppx/ppx.ml
davesnx: It might require a non-trivial refactor, but will do that I believe those locations are worth every detail

ppx/ppx.ml
```diff
@@ -572,10 +579,11 @@ let jsxMapper () =
           ; (* [|moreCreateElementCallsHere|] *)
             (nolabel, childrenExpr) ]
       | nonEmptyProps ->
-          (* Filtering out the props that don't have a label, not sure how's possible *)
-          let propsWithName =
-            List.filter (fun (name, _) -> getLabel name != "") nonEmptyProps
+          (* Filtering out last unit *)
+          let isLabeledArg (name, value) =
+            getLabel name != "" && not (isUnit value)
```
davesnx, line 584: Fixed the mistery @jchavarri

ppx/ppx.ml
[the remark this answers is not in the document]
davesnx: Made it more explicit, was the unit

ppx/ppx.ml
```diff
@@ -576,15 +576,32 @@ let jsxMapper () =
...
           in
+          let unionLocation first second =
+            let diff = Location.compare first second in
+            let loc_start = {first.loc_start with pos_lnum= diff} in
+            let loc_end = second.loc_end in
+            {loc_start; loc_end; loc_ghost= false}
+          in
           let makePropField (arg_label, value) =
+            let valueLoc = value.pexp_loc in
+            let argLabelLoc = unionLocation loc value.pexp_loc in
```
davesnx, line 595: I append both locations to ensure the argLabel is part of the error.

Before
```
File "example/src/Events.re", line 49, characters 6-13:
49 |       <button type_="submit" xxxx=0> {"submit dis" |> React.string} </button>
           ^^^^^^^
Error: prop 'xxxx' isn't a valid React prop
```

Passing the expression loc
```
49 |       <button type_="submit" xxxx=0> {"submit dis" |> React.string} </button>
                                       ^
Error: prop 'xxxx' isn't a valid React prop
```

After (with unioning the locations)
```
File "example/src/Events.re", lines 1-49, characters 6-35:
-1 |       <button type_="submit" xxxx=0> {"submit dis" |> React.string} </button>
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: prop 'xxxx' isn't a valid React prop
```

davesnx: Hope it's clear. We could get fancier: with the length of the argLabel and minus the expr loc minus 1 (the `=`) maybe can point to the argLabel. Let me know if you think is worth it.

jchavarri: This is great! Another idea (for the future ™️ ) could be to print after the error a link to a dedicated page in https://ml-in-barcelona.github.io/jsoo-react/ that explains native elements, and lists the whole list of tags that are allowed (the list could be shared code with the ppx ✨ ).

One nice thing is that because all these checks happen at compile time we don't need to care about the errors or urls making bundle size larger.

jchavarri: > I append both locations to ensure the argLabel is part of the error.

Would it be easier to pick the location from the parent ast node?

jchavarri: I mean, the `Pexp_apply` one:

image

davesnx: That's way smarter!

Pushed a commit with this change, and appended a small refactor afterwards.

jchavarri:
Bringing this in. Thanks again @davesnx !

PR #71, opened 2021-12-05
Type-safe HTML and SVG tags

davesnx:
# Description

This PR brings a layer of safety to writing React components that render HTML elements or SVG elements. The idea behind this isn't new, libraries like https://github.com/ocsigen/tyxml from OCaml, or https://github.com/kotlin/kotlinx.html from Kotlin Web, or https://github.com/nicojs/typed-html from TypeScript. 

Since this library/ppx focuses on React, it includes the divergences from HTML standard: https://reactjs.org/docs/dom-elements.html.

# Plan of action

1. Study how `tyxml` validate attributes: https://github.com/ocsigen/tyxml/blob/master/lib/html_types.mli
2. Check other implementations such as [types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts)
3. Estimate how much can take to re-implement the entire spec.

Tried to use tyxml to easily integrate with our codebase, but found that their implementation to derivate a bit from our needs.  The approach they took was more functional. Ensuring to provide a huge list of functions where those type-check against some nested open polyvars.

Our goal was to have a list of attributes and store this as metadata in order to generate the type-checks from the ppx. Using the types from types/react was easier to parse. I expect React's static types from TypeScript to be tested intensively and ensure we provide those React pieces that other HTML implementations will lack. That was the approach I took.

# Todo
- [x] Copy index.d.ts
- [x] Make it OCaml valid
- [x] Turn all attributes to String, comment the types (**1*)
- [x] Turn variants into let bindings to String (**1*)
- [x] Understand *`HTMLAttributes`
- [x] Uncomment `htmlElements`
- [x] Migrate Aria
- [x] Migrate SVG
- [x] Fix all `htmlNames`
- [x] Implement logic to generate type annotations on the ppx.ml
- [x] Spike events
- [x] Events in `commonDOMAttributes`
- [x]  Ensure attribute correctness
    - [x] Bring tyxml test cases
- [x] Few appearances of types that don't understand/don't used/aren't binded.
    - [x] `abstractView`
    - [x] `touch`
    - [x] `touchList`
    - [x] `errorInfo`
- [ ] ~~Transform lists to be Hash tables? any other structure that fits best?~~
    - [ ] ~~Any other pattern that might benefit us from the HTML metadata encoded?~~
- [x] Refactor names of the groups of attributes (domAttributes, reactAttributes, attributesHTML) Try to came up with the HTML Spec terms (add links to it)
- [ ] ~~Any way to battle-test it?~~ (Migrating real-world-example might make the trick)
- [ ] ~~Flag those that are experimental, or deprecated. Any way to run it automatically?~~ (Will be done in further PRs)
- [ ] ~~Improve error messaging~~ (Will be done in further PRs)
```ocaml
(* Errors look like this: *)

→ ~/C/g/r/jsoo-react on Type-safe-html-tags ✗ make

File "test.ml", lines 109-111, characters 2-10:
109 | ..((div ~cols:1 ~href:(("https://example.com")
110 |       [@reason.raw_literal "https://example.com"]) ~children:[foo] ())
111 |   [@JSX ])
Error: prop 'cols' isn't a valid prop for a 'div'

(* They should look like *)

File "test.ml", lines 109-111, characters 2-10:
109 | ..((div ~cols:1 ~href:(("https://example.com")
110 |       [@reason.raw_literal "https://example.com"]) ~children:[foo] ())
111 |   [@JSX ])
Error: 'div' contains an invalid prop: 'cols'.
It only accepts 'Global attributes' defined in https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
```

# Future

Note on (**1*): In this PR the goal wasn't to bring a nicer API to HTML/SVG elements, rather than validate its attributes. In further PRs we could bring those attributes that are straightforward to type in OCaml. Type-safe attributes might bring a lot of value to the table but they might have a cost where could be harder to learn (depending on the compiler messages), harder to document and harder to migrate into it.

# Discuss
- Do we want this?
- How errors should be displayed?
- Does it make sense to keep html.ml inside this repo?
- How can we document the types? I thought we could generate an .mli from the same metadata, but happy to discuss other approaches.
- How can we manage new additions/deletions to the spec? Flag them somehow?
- (as discussed in private, but happy to bring it again). Is there any way with ocaml.doc during the ppx generation where we could add descriptions? It will mean a lot of users could hover on the attribute and the description from MDN pops up.


davesnx:
Hey @dodomorandi, saw your PR adding polyvariants to rescript-react and your comments on the jsx ppx v4 from Ricky. This PR allows us to have fine granted type-safety for all elements and the next changes will include moving away from `strings` and introducing polyvariants or variants for many attributes. 

Would you like to participate and have a chat with me about this topic? I would love to know your opinions on the questions I ask on this PR. Let me know if you want to chat here or in discord, both work great for me :D

Hope this cc isn't to intrusive!

ppx/html.ml
```diff
@@ -15,558 +17,1859 @@ type eventType =
...
+
+let attributeAnchorTarget = String
+(* | Self
+   | Blank
+   | Parent
+   | Top
+   | Custom of String *)
+
+let domAttributes =
+  [ Event {name= "onCopy"; type_= Clipboard}
```
jchavarri, line 50: I wonder about OCaml syntax future support. Do you think we should use some live conversion from `name`, replacing capital letters with underscore? e.g. from `onCopy` to `on_copy`. One alternative could be to add another field `ocaml_name` to the record, but it seems to add a lot of boilerplate 😬

ppx/html.ml
```diff
@@ -15,558 +17,1859 @@ type eventType =
...
+  ; {tag= "u"; attributes= domAttributes & attributesHTML}
+  ; {tag= "ul"; attributes= domAttributes & attributesHTML}
+  ; {tag= "var"; attributes= domAttributes & attributesHTML}
+  ; {tag= "video"; attributes= domAttributes & videoHTMLAttributes}
+  ; {tag= "wbr"; attributes= domAttributes & attributesHTML}
+  ; { tag= "webview"
+    ; attributes= domAttributes & attributesHTML & webViewHTMLAttributes } ]
+
+let svgElements =
+  [ {tag= "svg"; attributes= domAttributes & svgAttributes & ariaAttributes}
```
jchavarri, line 1793: As all elements "inherit" from the same groups of attrs, might be worth having just a list of strings with the tags and a map function that builds the full record with `tag` and `attributes` fields.

ppx/html.ml
```diff
@@ -15,558 +17,1859 @@ type eventType =
...
+  ; Attribute {name= "partition"; htmlName= "partition"; type_= String}
+  ; Attribute {name= "plugins"; htmlName= "plugins"; type_= Bool}
+  ; Attribute {name= "preload"; htmlName= "preload"; type_= String}
+  ; Attribute {name= "src"; htmlName= "src"; type_= String}
+  ; Attribute {name= "useragent"; htmlName= "useragent"; type_= String}
+  ; Attribute {name= "webpreferences"; htmlName= "webpreferences"; type_= String}
+  ]
+
+(* Browser Interfaces https://github.com/nikeee/2048-typescript/blob/master/2048/js/touch.d.ts *)
+(* let abstractView = [] *)
```
jchavarri, line 1604: Why are these commented?

ppx/html.ml
```diff
@@ -15,558 +17,1859 @@ type eventType =
...
+  ; { tag= "input"
+    ; attributes= domAttributes & attributesHTML & inputHTMLAttributes }
+  ; {tag= "ins"; attributes= domAttributes & attributesHTML & insHTMLAttributes}
+  ; {tag= "kbd"; attributes= domAttributes & attributesHTML & domAttributes}
+  ; { tag= "keygen"
+    ; attributes= domAttributes & attributesHTML & keygenHTMLAttributes }
+  ; { tag= "label"
+    ; attributes= domAttributes & attributesHTML & labelHTMLAttributes }
+  ; {tag= "legend"; attributes= domAttributes & attributesHTML}
+  ; {tag= "li"; attributes= domAttributes & attributesHTML & liHTMLAttributes}
```
jchavarri, line 1713: Does it make sense to just inline `liHTMLAttributes` here? as they're only used once...

jchavarri (approved):
This looks amazing ❤️

ppx/html.ml
davesnx: If we want to provide both names, I would say to store it in one form and transform it during the ppx injection. WDYT?

ppx/html.ml
davesnx: Most attributes are moved directly from TypeScript. One of the lasts tasks I want to do is `[ ] Refactor names of the groups of attributes (domAttributes, reactAttributes, attributesHTML) Try to came up with the HTML Spec terms (add links to it)`

ppx/html.ml
davesnx: Those are artefacts from the copy, most of them are related with types on the event handlers for React. We already type those in events.mli

ppx/html.ml
```diff
@@ -15,558 +17,1762 @@ type eventType =
...
+      ; type_= String (* number |  *) }
+  ; Attribute
+      { name= "unicodeRange"
+      ; htmlName= "unicodeRange"
+      ; type_= String (* number |  *) }
+  ; Attribute
+      {name= "unitsPerEm"; htmlName= "unitsPerEm"; type_= String (* number |  *)}
+  ; Attribute
+      { name= "vAlphabetic"
+      ; htmlName= "vAlphabetic"
```
jchavarri, line 1499: Isn't this `v-alphabetic`? https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/v-alphabetic 

(also it's deprecated, not sure if it's worth including it)

ppx/html.ml
```diff
@@ -15,558 +17,1762 @@ type eventType =
...
+      ; type_= String (* number |  *) }
+  ; Attribute {name= "values"; htmlName= "values"; type_= String}
+  ; Attribute
+      { name= "vectorEffect"
+      ; htmlName= "vectorEffect"
```
jchavarri, line 1504: `vector-effect` https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/vector-effect

ppx/html.ml
```diff
@@ -15,558 +17,1762 @@ type eventType =
...
+      ; type_= String (* number |  *) }
+  ; Attribute {name= "version"; htmlName= "version"; type_= String}
+  ; Attribute
+      {name= "vertAdvY"; htmlName= "vertAdvY"; type_= String (* number |  *)}
```
jchavarri, line 1508: hm seems all svg attrs need revisiting https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/vert-adv-y

jchavarri (changes requested):
There seem to be some issues with svg attrs names.

-----

Comments on "Discuss":

> Do we want this?

I think so. I think one of the upsells of OCaml is to keep things type safe, so the more info we can provide to users about available attributes and values for them, the better.

> How errors should be displayed?

Current approach is ok for v1. For future versions, a couple of improvements that I can think are:
- Narrow down location to the actually labelled argument that is not recognized (instead of full pexp_apply node)
- Use some fuzzy search to include some potential candidates in error msg (the famous "did you mean `foo`?"

Does it make sense to keep html.ml inside this repo?

I think so. Unless there was a very clear flow to get the list of attributes from its original source. Btw, is the process for creation of the list of attributes and the original source for them documented anywhere in this PR or elsewhere?

---

Other questions:

- What happens with non-standard, or browser specific attributes? e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-slider-thumb
- This is just an idea re: errors, would be nice to add some test setup where we could test error messages from the ppx (and that chance to also assert locations)
- The `data-` attrs are for future PRs right?

Thanks again for working on this!

ppx/test/pp.expected
```diff
@@ -0,0 +1,736 @@
+module type Foo  =
```
jchavarri, line 1: This file should be removed (was moved to pp_reason.expected), prob a bad merge.

jchavarri (approved):
looks great, just removing `pp.expected` and :shipit:

PR #72, opened 2021-12-06
Better support for OCaml syntax

jchavarri:
This PR implements most of the ideas in #67.

- Moves calls to `createElement` inside the component implementation, to allow consumers to not use `JSX` attributes, just using plain functions, e.g. `let my_element = Greeting.make ~name:"John" ()`.
- The change above carries some constraints:
  - the ppx now keeps original Reason JSX ppx behavior, and wraps all children in a list
  - children types need to become of type `element list` to avoid warnings for keys missing in cases where it's not needed.
  - for components that pass `children` over, we need to re-enable spread operator `...`
- Removes the functionality that used to be needed for module signatures, as right now they can just be typed as regular functions with labelled args.
- Simplifies the handling of component implementation arguments in the ppx. In most of internal ppx code we don't care about `Nolabel` args (as only final `()` is allowed), so we create a new type with only `Labelled` and `Optional` values to remove handling edge cases that never happen.

## Todo

Maybe some (or all 😅 ) of these things can go in future PRs:

- [x] Add support for `let%component` syntax
- [x] Move all generated functions inside either an anonymous module (`open struct ... end`) or inside a top level binding `let make = let ....` to avoid them appearing externally
- [x] remove interface test in ppx, doesn't make sense now and adapt Interface text (to explain no action is needed)
- [ ] Add snake case version of all methods
- [ ] Call `setDisplayName` when building with `profile=dev` Dune mode to keep module names in React browser extension
- [ ] Better locations
- [ ] Make it so that `let make ~foo () bar =` fails to build


jchavarri:
Merging with the ideas in RFC implemented, can tackle remaining tasks in other PRs.

PR #73, opened 2021-12-06
Add 'help' command in Make

davesnx:
A tiny improvement to Make.

Let me know if make sense, (and if It doesn't break the make update deps automatically. )


jchavarri (approved):
thanks!

PR #87, opened 2021-12-14
Create a test case for capturing error messages at build(ppx)-time

davesnx:
This test ensures that our error remains the same during new iterations. The test ensures that the ppx checks the validity of HTML attributes based on the HTML tag, since these operations happen at build-time we can run the ppx and capture it's stderr output and diff it against a snapshot.

With this method, we can't ensure that the type-checker works thought. There are methods to run the type-checker for example I'm doing it in styled-ppx https://github.com/davesnx/styled-ppx/blob/main/packages/ppx/test/native/Static_test.re and https://github.com/davesnx/styled-ppx/blob/main/packages/ppx/test/native/dune. We could look into those closely.

The PR merges all runtest into one action, to review the capturing error check: [`e549642` (#87)](https://github.com/ml-in-barcelona/jsoo-react/pull/87/commits/e549642d31a6459f728ce9a7620a0770d090bd2e)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,3 +55,7 @@ jobs:
 
       - name: Run tests
         run: make test
+
+      - name: Run error message test
+        if: ${{ matrix.os != 'windows-latest' }}
+        run: make test-error-msg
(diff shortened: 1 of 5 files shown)


jchavarri (approved):
Looks awesome 🚀

PR #89, opened 2021-12-20
Remove filtering of optional props in make_js_props_obj

### LINKED ISSUE #83: Remove filtering of undefined props passed to upper case components (except `key`)
There's no need to filter optional props on uppercase components when calling `createElement`, they can (and should) be passed as is. This will allow to get rid of array->list->array conversion and filtering on each element creation call.

The only exception is `key` property, which is handled by React.

Questions:
- What about `ref`?

jchavarri:
Fixes #83.

Optional props passed to uppercase components should be left as they are, no conversion to `Js.Optdef.t` required. The only exceptions are `key` and `ref` which are read by React.js.

Previous implementation led to bugs when using optional prop + default value, where the component would receive `undefined` when some value was passed (see added tests in  test suite).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ test: ## Run the unit tests
 	$(DUNE) build @runtest
 
 test-promote: ## Updates snapshots and promotes it to correct
-	$(DUNE) build @runtest
+	$(DUNE) build @runtest --auto-promote
 
 # This is a separate command to run this tests conditionaly on CI, currently the dune setup with (bash "! ./%{main} %{input}")) fails in Windows. We skip it on Windows.
 test-error-msg: ## Run the unit tests for error messages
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -26,7 +26,7 @@ test: ## Run the unit tests
 	$(DUNE) build @runtest
 
 test-promote: ## Updates snapshots and promotes it to correct
-	$(DUNE) build @runtest
+	$(DUNE) build @runtest --auto-promote
```
davesnx, line 29: my bad 😞

davesnx (approved):
That's a great improvement.

jchavarri:
> That's a great improvement.

Not possible without your help + work on lowercase comps optimizations! 🙂

Issue #81, opened 2021-12-11
Can't pass optional values to dom elements


jchavarri:
Repro:

```ocaml
[@@@react.dom]

let href = None in
let element = a ?href ()
(*               ^^^^
This expression has type string option but an expression was expected of type
  string *)
```

davesnx:
Ops, that's because the type is enforced by HTML. Should we allow to be wrapped with option?

It's relatively easy to do!

jchavarri:
I think it'd be useful :) The sample code I was working on had either `href` or `onClick` depending on some optional value. We can work around it by building two elements, one with `href` and one with `onClick`, but it's kind of sad to not be able to leverage optional labelled args, no? (if you say it's easy...)

jchavarri:
@davesnx it seems this is still not possible.

Input:

```ocaml
let element = a ?href ()
              ^^^^^^^^^^
Error: This expression has type string option
       but an expression was expected of type string
make: *** [build] Error 1
```

If I expand the expression:

```ocaml
let element =
  React.Dom.createDOMElementVariadic "a"
    ~props:(Js_of_ocaml.Js.Unsafe.obj
              [|("href",
                  (Js_of_ocaml.Js.Unsafe.inject
                     (Js_of_ocaml.Js.Optdef.option
                         (Js_of_ocaml.Js.string (href : string option)))))|] :
                                                ^^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type string option
       but an expression was expected of type string
make: *** [build] Error 1
    React.Dom.domProps) []
```

(we should prob add a test to `test_jsoo_react` suite)

jchavarri:
This was fixed in https://github.com/ml-in-barcelona/jsoo-react/pull/92. Thanks @davesnx !

PR #92, opened 2021-12-29
Allow `string option` transformations for native elements

davesnx:
This PR fixes a remaining bug for optional parameters, from https://github.com/ml-in-barcelona/jsoo-react/issues/81#issuecomment-1002261158

Was introduced by https://github.com/ml-in-barcelona/jsoo-react/pull/84.

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -440,7 +440,7 @@ let rec makeFunsForMakePropsBody list args =
 let makeAttributeValue ~loc ~isOptional (type_ : Html.attributeType) value =
   match (type_, isOptional) with
   | String, true ->
-      [%expr Js_of_ocaml.Js.string ([%e value] : string option)]
+      [%expr Option.map Js_of_ocaml.Js.string ([%e value] : string option)]
   | String, false ->
       [%expr Js_of_ocaml.Js.string ([%e value] : string)]
   | Int, false ->
--- a/ppx/test/input_ocaml.ml
+++ b/ppx/test/input_ocaml.ml
@@ -17,3 +17,5 @@ let%component make ~children:(first, second) () =
   div ~children:[first; second] ()
 
 let%component make ?(name = "") = div ~children:[name] ()
+
+let%component make () = a ?href:(Some "https://opam.ocaml.org") ()
--- a/test/test_jsoo_react.re
+++ b/test/test_jsoo_react.re
@@ -92,6 +92,32 @@ let testOptionalProps = () => {
   });
 };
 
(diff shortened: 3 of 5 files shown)


ppx/test/input_ocaml.ml
```diff
@@ -17,3 +17,5 @@ let%component make ~children:(first, second) () =
   div ~children:[first; second] ()
 
 let%component make ?(name = "") = div ~children:[name] ()
+
+let%component make () = a ?href:(Some "https://opam.ocaml.org") ()
```
jchavarri, line 21: Could we add a test in `test_jsoo_react` please? ppx tests will swallow all the things as long as they're syntactically valid, so they don't allow to detect type errors.

davesnx: I wanted to add them as example, but forgot we have tests that run the browser 💯 

Fixed in [`7c03b8b` (#92)](https://github.com/ml-in-barcelona/jsoo-react/pull/92/commits/7c03b8bf96a79f964dabf582fecf3e38c05f3b38)

test/test_jsoo_react.re
```diff
@@ -92,6 +92,32 @@ let testOptionalProps = () => {
   });
 };
 
+let testOptionalAttributes = () => {
```
jchavarri, line 95: Maybe we could call this test `testOptionalPropsLowercase` and rename the other one to `testOptionalPropsUppercase`?

ppx/ppx.ml
```diff
@@ -440,7 +440,7 @@ let rec makeFunsForMakePropsBody list args =
 let makeAttributeValue ~loc ~isOptional (type_ : Html.attributeType) value =
   match (type_, isOptional) with
   | String, true ->
-      [%expr Js_of_ocaml.Js.string ([%e value] : string option)]
+      [%expr Option.map Js_of_ocaml.Js.string ([%e value] : string option)]
```
jchavarri, line 443: This does not affect any other cases because we only convert the values for the `String` case right?

jchavarri (approved):
Looks great! thanks for the fix + adding tests 🙏

ppx/ppx.ml
davesnx: Exactly, any other case is handled with OptDef.

zbaylin:
Just want to chime in and say thanks for this @davesnx ❤️! I bumped up against this when trying to create components that used `input`, i.e. wanting to only set `value` if a value for it is passed in above. Will be nice to have this in 🎉

PR #96, opened 2021-12-29
refactor: require `react`, `react-dom` within OCaml

zbaylin:
This removes the need to set React globally using a file similar to `examples/src/react-requires.js`. 

Rather than `gen_js_api` referring to a global named `React`/`ReactDOM`, we import the libraries once using `require` and partially apply `js.call`s

Thanks @jchavarri for the help with this!


README.md
```diff
@@ -46,42 +46,11 @@ For new projects, the best way to start is by using [Spin](https://github.com/tm
...
-      (javascript_files react-requires.js)))
-    ```
-
-    To see an example of this approach, check the [example](example) folder.
+    
+    `jsoo-react` uses `require` to import React and ReactDOM. This means that you will likely need to use a bundler such as Webpack or rollup.js.
 
     Note that at this moment, `jsoo-react` is compatible with **React 16**, so be sure to have the appropriate constraints in your `package.json`.
 
-    #### With `<script>` tags
```
jchavarri, line 74: Hm right, it's a bit of a pity that we lose this possibility. But I guess it's fine? The only app out there that I know of uses webpack: https://github.com/michael-schwarz/goblint-webfrontend-jsoo/blob/2d88e9649078d0138ff5f1e76e2391f7cd684e01/gobview/webpack.config.js

cc @davesnx @michael-schwarz @keremc

lib/core.mli
```diff
@@ -1,4 +1,4 @@
-type element = private Ojs.t
```
jchavarri, line 1: If we remove this we "leak" the internal implementation of `element` and stops being abstract, right? I'm not sure we want to do this.

ppx/ppx.ml
```diff
@@ -653,13 +653,13 @@ let rec recursivelyTransformNamedArgsForMake mapper ctxt expr list =
             None
       in
       recursivelyTransformNamedArgsForMake mapper ctxt expression
-        ( ( Str_label.of_arg_label arg
```
jchavarri, line 656: I think this is what made ci fail, did you install latest `ocamlformat` with `opam install ocamlformat`?

jchavarri (commented):
Looks awesome! thanks a lot for working on this.

I think once the ci issue is fixed, this would be good to merge imo.

lib/core.mli
zbaylin: Unfortunately, the way the ppx works with `Ojs.call` means that the input and output types both have to be the same, i.e.

```ocaml
external call: t -> string -> t array -> t = "caml_js_meth_call"
```

This means that `Imports.react` and `element` have to be the same type. There might be a way to get around this that I can look into, though

ppx/ppx.ml
zbaylin: Yeah that was my bad, I was using an outdated version of `ocamlformat` -- should be fixed now :)

lib/core.mli
jchavarri: Hm I see. I'm not sure exactly how, but if I restore the `private Ojs.t` definition, everything seems to keep working. 😕

jchavarri: (to clarify, I would expect `Core` compilation to work fine as the internal implementation is visible inside the same module, but I would expect usages in `Dom` to fail, as they can't see the internal implementation of `element`)

zbaylin: Oh you're right, that's very interesting. It might have been the way I was writing the bindings previously that produced the type error, and this way is fine (I'm also not sure why that would be, but `gen_js_api` uses `Obj.magic` in a lot of places, so maybe that's it)

README.md
zbaylin: Yeah that's what I meant yesterday when I said this would be forcing people to use a bundler, but I don't think it's a horrible trade-off as most people bundle React that way. 

The one way to get around that would be to functorize the interface that produces the React bindings, but I think that might require more refactoring for another PR.

jchavarri: I agree. I wouldn't add the functor though until there's someone with a use case that requires it. :)

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-compiler:
-          - 4.10.2
+          - 4.12.1
```
jchavarri, line 28: @zbaylin Was this to work around the Windows issues? I'd rather move the update to newer compiler to a separate PR if possible (I think we should prob add some kind of cross-version builds + tests in ci, but that's a larger / different topic).

zbaylin: Yeah that was to see if the OCaml version was causing the CI failure -- I'll change it back now :)

zbaylin: Reverted in 95db987 😃

jchavarri:
Thanks again @zbaylin for working on this! 🎸 Merging

README.md
michael-schwarz: Yup, we indeed use webpack. 
It's really exciting to see all the recent work y'all have been doing with `jsoo-react`, thank you! :tada: 

(Btw, we once again moved the project to a new (now permanent) repo within our GitHub organization https://github.com/goblint/gobview)

PR #101, opened 2022-01-02
Keep useReducer dispatch and useState updater refs constant

### LINKED ISSUE #100: New updater functions being returned by useState on every render
The generated code for `useState` is:

```js
function useState(x278) {
  var
  ___ =
    caml_js_wrap_callback_strict
      (1, function (param) { return caml_call1(x278, 0) }),
  x279 = Ojs[18].React.useState(___);
  function _$_(x280) {
    var
    _aa_ =
      caml_js_wrap_callback_strict
        (1, function (x281) { return caml_call1(x280, x281) });
    caml_call2(Ojs[16], x279, 1)(_aa_);
    return 0
  }
  return [0, caml_call2(Ojs[16], x279, 0), _$_]
}
```

This is problematic because every time a new function `_$_` is created, which breaks the assumptions of React, where if you do
```ocaml
let data, setData = React.useState (fun () -> 2)
```

`setData` should remain constant over time, but I am not sure it's possible as it'd require removing `caml_js_wrap_callback_strict`. I'm not sure why wrapping is needed for functions that only take 1 param 🤔

jchavarri:
Fixes #100.

The fix consist mostly on inlining the previously generated code by gen_js_api and removing the wrapping over the following functions:
- The updater returned in the 2nd position of the tuple in `useState`
- The dispatch returned in 2nd position of the tuple in `useReducer`

Both functions can take only 1 argument, so I don't see any reason why the currying handling is needed in these case.

Added some tests for this as well as a base test for `useState`, which was not existent.

### DIFF
--- a/lib/core.mli
+++ b/lib/core.mli
@@ -522,21 +522,40 @@ val useMemo7 : (unit -> 'value) -> 'a * 'b * 'c * 'd * 'e * 'f * 'g -> 'value
 
 val useState : (unit -> 'state) -> 'state * (('state -> 'state) -> unit)
   [@@js.custom
-    val useState_internal :
-      Imports.react -> (unit -> 'state) -> 'state * (('state -> 'state) -> unit)
-      [@@js.call "useState"]
+    let (useState_internal :
+             Imports.react
+          -> (unit -> 'state)
+          -> 'state * (('state -> 'state) -> unit) ) =
+     fun (react : Imports.react) (init : unit -> 'state) ->
+      let result =
+        Ojs.call
+          (Imports.react_to_js react)
+          "useState"
+          [|Ojs.fun_to_js 1 (fun _ -> Obj.magic (init ()))|]
+      in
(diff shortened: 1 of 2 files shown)


jchavarri:
Here's an example of the transformation:

```diff
let (useState_internal :
      Imports.react -> (unit -> 'state) -> 'state * (('state -> 'state) -> unit)
      ) =
 fun (react : Imports.react) (init : unit -> 'state) ->
  let result =
    Ojs.call
      (Imports.react_to_js react)
      "useState"
      [|Ojs.fun_to_js 1 (fun _ -> Obj.magic (init ()))|]
  in
  ( Obj.magic (Ojs.array_get result 0)
- , fun (updater : 'state -> 'state) ->
-     ignore
-       (Ojs.apply (Ojs.array_get result 1)
-          [| Ojs.fun_to_js 1 (fun (data : Ojs.t) ->
-                 Obj.magic (updater (Obj.magic data)) ) |] ) )
+ Obj.magic (Ojs.array_get result 1))

let useState initial = useState_internal Imports.react initial
```

PR #103, opened 2022-01-04
chore(deps): open jsoo constraint to 3.11.0

zbaylin:
It doesn't seem as if jsoo uses semver in the conventional way as 3.11.0 [doesn't contain any breaking changes](https://github.com/ocsigen/js_of_ocaml/blob/master/CHANGES.md#3110-2021-10-06---lille). 

I was having some versioning conflicts in a project I was trying to use jsoo-react on, this should help fix that :)


jchavarri:
Thanks @zbaylin !

PR #107, opened 2022-01-08
fix build & test instructions in CONTRIBUTING

glennsl:
- `opam install ...` is already done by `make init`
- `make` does nothing, do `make build` instead
- JS dependencies in `test` must be installed before running `make test`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,9 +4,10 @@
 
 - Install opam: https://opam.ocaml.org/doc/Install.html
 - Create local switch by running `make init` (this step will take a few minutes, as it will build whole OCaml compiler)
-- Install deps with `opam install . --deps-only --with-test`
 - Build with `make`
-- Run tests with `make test`
+- Run tests with 
+  - Install test dependencies with `(cd test; yarn)`
+  - `make test`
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,9 +4,10 @@
 
 - Install opam: https://opam.ocaml.org/doc/Install.html
 - Create local switch by running `make init` (this step will take a few minutes, as it will build whole OCaml compiler)
-- Install deps with `opam install . --deps-only --with-test`
-- Build with `make`
-- Run tests with `make test`
+- Build with `make build`
+- Run tests with 
+  - Install test dependencies with `(cd test; yarn)`
```
jchavarri, line 9: I wonder if we should just run `yarn` from `make`, so there's only a single entry point for commands cc @davesnx

jchavarri (approved):
Thanks!

CONTRIBUTING.md
glennsl: From `make init` you mean? I think that makes sense a least!

davesnx: Yes, make should be the only tool you need to interfere to develop

PR #110, opened 2022-01-11
ppx: fix inexistent locations on comps that return element list

### LINKED ISSUE #90: Inexistent locations on error when component returns list of elements
```ocaml
let%component make () = [ React.null ]
```

Will trigger this error:

```
Error: This expression has type
         <  > Js_of_ocaml.Js.t -> React__Core.element list
       but an expression was expected of type
         <  > Js_of_ocaml.Js.t React.component =
           <  > Js_of_ocaml.Js.t -> React__Core.element
       Type React__Core.element list is not compatible with type
         React__Core.element
```

but with no locations.

jchavarri:
Fixes #90.

Instead of threading locations to the `make` function that is ultimately passed to `createElement` (which is what was originally triggering the error) we wrap the result of "ml_component" (the implementation provided by user) with `: React.element` annotation.

Then, the error happens earlier, in code that is not generated by ppx but written by user.

Tested manually (idk how to test ppx with full typed compilation 😅 ):

```
File "example/src/HelloWorldOCaml.ml", line 5, characters 26-38:
5 | let%component make ~foo = [React.null]
                              ^^^^^^^^^^^^
Error: This expression has type 'a list
       but an expression was expected of type React.element
```

```
File "example/src/HelloWorldOCaml.ml", line 5, characters 24-36:
5 | let%component make () = [React.null]
                            ^^^^^^^^^^^^
Error: This expression has type 'a list
       but an expression was expected of type React.element
```

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -899,18 +899,38 @@ let process_value_binding ~pstr_loc ~inside_component ~mapper ~ctxt binding =
         | { pexp_desc=
               Pexp_fun
                 ( Nolabel
-                , _default
-                , {ppat_desc= Ppat_construct ({txt= Lident "()"}, _) | Ppat_any}
-                , _internalExpression ) } ->
-            ((fun a -> a), true, expression)
+                , default
+                , ( { ppat_desc=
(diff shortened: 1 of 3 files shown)


davesnx (approved):
Great improvement!

PR #115, opened 2022-01-13
PoC: Dom lite

jchavarri:
This is a proof of concept for a potential API for OCaml that does not rely on any preprocessing at all for element creation.

Triggered by discussion from @glennsl in https://github.com/ml-in-barcelona/jsoo-react/discussions/113.

### DIFF
--- a/lib/dom_lite.ml
+++ b/lib/dom_lite.ml
@@ -1,0 +1,31 @@
+module Fragment = struct
+  include Core.Fragment
+
+  let make children = make ~children ()
+end
+
+let href str = ("href", str)
+
+let target str = ("target", str)
+
+let className str = ("className", str)
+
+let make_props list : Dom.domProps =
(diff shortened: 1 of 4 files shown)


lib/dom_lite.ml
```diff
@@ -0,0 +1,31 @@
...
+
+let target str = ("target", str)
+
+let className str = ("className", str)
+
+let make_props list : Dom.domProps =
+  Js_of_ocaml.Js.Unsafe.obj
+    ( Array.of_list list
+    |> Array.map (fun (key, value) ->
+           (key, Js_of_ocaml.Js.Unsafe.inject (Js_of_ocaml.Js.string value)) )
```
glennsl, line 17: We can skip this `Array.map` by doing this in each attribute constructor function instead.

jchavarri:
This was implemented fully by @glennsl in https://github.com/ml-in-barcelona/jsoo-react/pull/119.

PR #116, opened 2022-01-19
style: use functions as style key creation mechanism

### LINKED ISSUE #112: Make React.Dom.Style.make more bundle size friendly
Right now, every call to `React.Dom.Style.make` is compiled to a bunch of zeros on unused optional arguments.

This leads quickly to larger bundle sizes, e.g. for 7 usages of this function in some real world project, a prod bundle increases by ~17KB.

We should provide an API where the style properties are passed in a list, so unused ones don't lead into bundle costs.

e.g. like @davesnx does in [jsoo-css](https://github.com/ml-in-barcelona/jsoo-css):

```reason
  let box =
    React.Dom.Style.make([|
      color("333"),
      backgroundColor("C0FFEE"),
      padding("45px"),
      borderRadius("8px"),
      fontSize("22px"),
      fontWeight("bold"),
    |]);
```

I don't think we need type safety on each prop type, for a first version.

jchavarri:
Fixes #112.

Just as an example, in the prod bundle of `example` demo app, size goes from 115KB to 103KB 😅 

Based on the ideas discussed in #113, and @glennsl [additions to poc](https://github.com/glennsl/jsoo-react/blob/4d42d539c66b035bcb4ddba37cf4245db21f38bf/lib/dom_lite.ml).

### DIFF
--- a/lib/dom.mli
+++ b/lib/dom.mli
@@ -46,8 +46,6 @@ val renderToElementWithId : Core.element -> string -> unit
       | Some element ->
           render reactElement element]
 
-type style
-
 type domRef = private Ojs.t
 
 module Ref : sig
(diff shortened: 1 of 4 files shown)


davesnx (approved):
Love the approach. Improves the API, using an array is substantially better than labelled arguments for this use-case. No need to bring functions such as merge styles or combine.

Reduces the bundle size by removing the 0s, but does it add all these functions into the resultant bundle?

lib/dom.mli
```diff
@@ -103,426 +103,1403 @@ val forwardRef : ('props -> domRef -> Core.element) -> 'props Core.component
     let forwardRef renderFunc = forwardRef_internal Imports.react renderFunc]
 
 module Style : sig
+  [@@@js.stop]
+
+  type style_key
+
   type t = style
```
glennsl, line 110: If we want to use proper CSS terminology, `style_key` would be a "declaration" ("decl" for short), and `style` would be a "declaration block".

lib/dom.mli
```diff
@@ -103,426 +103,1403 @@ val forwardRef : ('props -> domRef -> Core.element) -> 'props Core.component
...
+  val rubyPosition : string -> style_key
+
+  [@@@js.start]
+
+  [@@@js.implem
+  type style_key = string * Js_of_ocaml.Js.Unsafe.any
+
+  type t = style
+
+  let string_style_key key value =
```
glennsl, line 805: And the "key" part of the declaration is called a "property", so it would perhaps be more accurate to call this `string_style_prop`, at least in the way it's being used, partially applied.

jchavarri:
@glennsl @davesnx thanks for the review :) I updated the names to be closer to CSS semantics.

> but does it add all these functions into the resultant bundle?

No, only what's used. js_of_ocaml is quite efficient in terms of dead-code optimization.

PR #118, opened 2022-01-21
fix: Add location on errors for invalid key/ref usage on props

davesnx:
Added locations on errors related with `key` and `ref` used on props. 

I didn't add a test's error for those ones, I'm not sure If we want to get a test case for each error message we trigger. Eventually yes, but our system makes adding a ton of files to achieve that. WDYT?


jchavarri (approved):
🎉

PR #124, opened 2022-01-26
ppx: don't assume external components are functions

### LINKED ISSUE #123: Fix external%component to not make assumptions about the JS component type
Reported by @glennsl  in https://github.com/ml-in-barcelona/jsoo-react/pull/106#issuecomment-1021597956.

Current code generated by ppx for `external%component` adds a `make` function with arity 1 that just makes a function call to the JS component and pass props. But JS components can be objects too (`forwardRef`, class components and such).

Ppx should not make any assumptions about JS components.

We might work around this by making that the ppx output code something like this:

```diff
-let make
-  (Props :
-    < name: Js.js_string Js.t Js_of_ocaml.Js.readonly_prop   > 
-      Js_of_ocaml.Js.t)
-  =
-  (((Js_of_ocaml.Js.Unsafe.js_expr
-        "require(\"my-react-library\").MyReactComponent") Props)
-  [@warning "-20"]) in
fun ~name ->
-  fun ?key -> fun () -> React.createElement make (make_props ?key ~name ())
+  let t ?key () =
+    React.createElement
+      (Js_of_ocaml.Js.Unsafe.js_expr "require(\"my-react-library\").MyReactComponent" )
+      (make_props ?key ~name ())
```

So we just remove any of the assumptions and let React handle the component itself, directly.

glennsl:
Fixes #123

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -1215,48 +1215,22 @@ let jsxMapper () =
             let empty_loc = Location.in_file filename in
             let binding_pat_loc = empty_loc in
             let outer_make expression =
-              let react_component_attr =
-                try Some (List.find hasAttr pval_attributes)
-                with Not_found -> None
-              in
-              let payload =
-                match react_component_attr with
-                | Some {attr_payload} ->
-                    Some attr_payload
-                | None ->
-                    None
-              in
-              let make_js_comp ~loc ~fn_name ~named_type_list rest =
-                let props = get_props_attr payload in
+              let make_js_comp ~loc ~fn_name rest =
                 let inner_expr =
                   [%expr
-                    (Js_of_ocaml.Js.Unsafe.js_expr
-                       [%e constantString ~loc pval_prim] )
-                      [%e Exp.ident ~loc {txt= Lident props.propsName; loc}]
-                    [@warning "-20"]]
+                    Js_of_ocaml.Js.Unsafe.js_expr
+                      [%e constantString ~loc pval_prim]]
                 in
                 Exp.mk ~loc
                   (Pexp_let
                      ( Nonrecursive
-                     , [ Vb.mk
-                           (Pat.var {loc; txt= fn_name})
-                           (Exp.fun_ nolabel None
-                              { ppat_desc=
-                                  Ppat_constraint
-                                    ( makePropsName ~loc props.propsName
-                                    , makePropsType ~loc named_type_list )
-                              ; ppat_loc= loc
-                              ; ppat_attributes= []
-                              ; ppat_loc_stack= [] }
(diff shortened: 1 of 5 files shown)


ppx/ppx.ml
```diff
@@ -1285,7 +1259,7 @@ let jsxMapper () =
                                     , Exp.construct {loc; txt= Lident "()"} None
                                     ) ] )]] )
                  in
-                 make_props @@ js_comp @@ outer )
+                 make_props (js_comp outer) )
             in
             let inner_make_ident =
```
jchavarri, line 1264: To keep things working, instead of passing this `inner_make_ident` to `outer_make`, we would pass `js_component_expr`.

Something like:

```ocaml
            let js_component_expr =
              (* I need to revisit how all locations are created... 🙈  *)
              let loc = pstr_loc in
              [%expr
                Js_of_ocaml.Js.Unsafe.js_expr [%e constantString ~loc pval_prim]]
            in
            { pstr_loc
            ; pstr_desc=
                Pstr_value (Nonrecursive, [outer_make js_component_expr])
            }
            :: returnStructures ) )
```

ppx/ppx.ml
```diff
@@ -1215,48 +1215,22 @@ let jsxMapper () =
...
-              let payload =
-                match react_component_attr with
-                | Some {attr_payload} ->
-                    Some attr_payload
-                | None ->
-                    None
-              in
-              let make_js_comp ~loc ~fn_name ~named_type_list rest =
-                let props = get_props_attr payload in
+              let make_js_comp ~loc ~fn_name rest =
```
jchavarri, line 1218: I believe we can remove `make_js_comp` altogether to make the code simpler, as it brings no value anymore.

test/test_jsoo_react.re
```diff
@@ -890,6 +890,40 @@ let testExternals = () => {
...
+    });
+    assert_equal(c##.innerHTML, Js.string("<span>Hey <em>John</em></span>"));
+  });
+};
+
+let testExternalNonFunction = () => {
+  module JsComp = {
+    [@react.component]
+    external make: (~name: Js.t(Js.js_string)) => React.element =
+      {|require("./external").NonFunctionGreeting|};
```
jchavarri, line 917: 💙

jchavarri (approved):
This looks great! I have just a small suggestion to remove a bit more code if you feel up to it :) 🧹 

Otherwise I can merge and we'll do it later on.

glennsl:
There. I did try to do it like this initially, but couldn't quite figure out how things fit together. Thanks for the pointers!

jchavarri (approved):
Thank you so much!

PR #126, opened 2022-01-27
ppx: fix external optional args

### LINKED ISSUE #125: external%component props from optional arguments are wrapped in option
There seems to be no special handling of optional arguments for `external%component`, which means that optional arguments are passed in as options.

Given

```ml
external%component my_component ?num_leprechauns:int -> React.element = "..."
```

`my_component ()` will cause `0` to be passed as `num_leprechauns`.

`my_component ~num_leprechauns:42 ()` will cause `[0, 42]` to be passed as `num_leprechauns`.

It might be sufficient to convert the `option` value to a `Js.optdef` value, but I think the proper solution is to omit it from the props object if the argument is not passed. I'll try doing the simpler thing for now though.

glennsl:
Fixes #125

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -544,21 +544,21 @@ let makeValue ~loc ~isOptional prop value =
 
 let make_js_props_obj ~loc named_arg_list_with_key_and_ref =
   let label_to_tuple label =
-    let l = Str_label.str label in
-    let id = Exp.ident ~loc {txt= Lident l; loc} in
-    match l with
+    let label_str = Str_label.str label in
+    let id = Exp.ident ~loc {txt= Lident label_str; loc} in
+    match label_str with
     | "key" ->
         [%expr
-          [%e Exp.constant ~loc (Const.string l)]
+          [%e Exp.constant ~loc (Const.string label_str)]
           , inject
               (Js_of_ocaml.Js.Optdef.option
                  (Option.map Js_of_ocaml.Js.string [%e id]) )]
     | "ref" ->
         [%expr
-          [%e Exp.constant ~loc (Const.string l)]
+          [%e Exp.constant ~loc (Const.string label_str)]
           , inject (Js_of_ocaml.Js.Optdef.option [%e id])]
-    | l ->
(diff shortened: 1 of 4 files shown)


ppx/ppx.ml
```diff
@@ -544,20 +544,46 @@ let makeValue ~loc ~isOptional prop value =
...
+  let label_to_tuple label =
+    let label_str = Str_label.str label in
+    let id = Exp.ident ~loc {txt= Lident label_str; loc} in
+    match label with
+    | Optional "key" ->
+        [%expr
+          [%e Exp.constant ~loc (Const.string label_str)]
+          , inject
+              (Js_of_ocaml.Js.Optdef.option
+                 (Option.map Js_of_ocaml.Js.string [%e id]) )]
```
jchavarri, line 581: I am thinking now this is a bit confusing. In most cases at the moment we assume there is no conversion of types in the external definition, but for `key` we will convert from `string` to `js_string`.

I'm not sure what we should do yet, as it's unclear to me if we have to go down the conversion path as you suggest, or leave types untouched. But whatever we do, I think it should be consistent 🤔

test/test_jsoo_react.re
```diff
@@ -1058,6 +1072,7 @@ let externals =
     "basic" >:: testExternals,
     "children" >:: testExternalChildren,
     "non-function" >:: testExternalNonFunction,
+    "optioal-arg" >:: testExternalOptionalArg,
```
jchavarri, line 1075:
```suggestion
    "optional-arg" >:: testExternalOptionalArg,
```

jchavarri (approved):
Thank you! I am merging this for now. I had some hesitation about how to consistently convert types in `external` definitions, but I am also noticing your other PR #127 which does convert between all types. I will try to review that other one soon.

test/test_jsoo_react.re
jchavarri: Fixed in https://github.com/ml-in-barcelona/jsoo-react/commit/347b4ef114dd8d2a789e0248005e4461b1c9e9e7

PR #127, opened 2022-01-29
PoC: automatic js ffi conversion of external props

glennsl:
For the purpose of discussion first of all, this is a proof of concept of one approach to add support for automatic conversion of a select few native types to their js equivalents for props on external component definitions (as briefly discussed [here](https://github.com/ml-in-barcelona/jsoo-react/pull/106#issuecomment-1021309799)).

The currently implemented conversions are:

- `string` -> `Js.js_string Js.t`
- `bool` -> `bool Js.t`
- `'a array` -> `'a Js.js_array Js.t` where `'a` will also be recursively converted
- `React.element list` -> `React.element Js.js_array Js.t` (in order for JSX children to work as expected, unsure if it should convert lists in general)

Everything else will be passed on as-is, which means you can still use `Js.js_string Js.t` directly, for example, and have it work as expected.

Beware, however, that this is just a syntactic rewrite. It doesn't actually know what the types are, so if they're aliased, shadowed or qualified in some unexpected way, this will break. It's very unlikely that this will happen with such fundamental types though. And in most cases you should get reasonable errors. Because the wrapping is done using non-polymorphic functions it will tell you exactly what it expects. The cases where it doesn't is if you expect conversion, but it doesn't, since then it'll just pass the value on to JavaScript where it will result in strange things happening, as is the norm over in untyped JS-land.

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -593,6 +593,16 @@ let make_external_js_props_obj ~loc named_arg_list =
           , inject
               (Js_of_ocaml.Js.Optdef.option
                  (Option.map Js_of_ocaml.Js.bool [%e id]) )]
+    | Labelled l, Some [%type: [%t? _] array] ->
+        [%expr
+          [%e Exp.constant ~loc (Const.string l)]
+          , inject (Js_of_ocaml.Js.array [%e id])]
+    | Optional l, Some [%type: [%t? _] array] ->
+        [%expr
+          [%e Exp.constant ~loc (Const.string l)]
+          , inject
+              (Js_of_ocaml.Js.Optdef.option
+                 (Option.map Js_of_ocaml.Js.array [%e id]) )]
     | Labelled l, _ ->
         [%expr [%e Exp.constant ~loc (Const.string l)], inject [%e id]]
     | Optional l, _ ->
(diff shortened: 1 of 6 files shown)


ppx/ppx.ml
```diff
@@ -568,9 +568,56 @@ let make_js_props_obj ~loc named_arg_list_with_key_and_ref =
...
+      Some [%expr Js_of_ocaml.Js.bool]
+  | Some [%type: [%t? inner_typ] array] ->
+      Some
+        ( match get_wrap_fn_for_type ~loc (Some inner_typ) with
+        | Some inner_wrapf ->
+            [%expr fun v -> Js_of_ocaml.Js.array (Array.map [%e inner_wrapf] v)]
+        | None ->
+            [%expr Js_of_ocaml.Js.array] )
+  | _ ->
+      None
```
glennsl, line 587: This is essentially all that's needed to support these conversions.

jchavarri (approved):
This is really cool 💯 

> (in order for JSX children to work as expected, unsure if it should convert lists in general)

I don't think we should convert lists in general, as it's not a type that is considered as convertible to JS, generally, in jsoo [docs](https://ocsigen.org/js_of_ocaml/latest/manual/library) or code.

> Beware, however, that this is just a syntactic rewrite. It doesn't actually know what the types are, so if they're aliased, shadowed or qualified in some unexpected way, this will break. It's very unlikely that this will happen with such fundamental types though. And in most cases you should get reasonable errors.

I think this sounds reasonable, we can test how this goes.

Thanks!

ppx/ppx.ml
davesnx: That's pretty clever, massive job @glennsl !

PR #128, opened 2022-01-29
ppx: Remove @@@react.dom + transform lowercase elements to Dsl functions

jchavarri:
Continues integration work after @glennsl work in #119.

- Update ppx to generate fun application calls to element names, e.g. `<div/>` is transformed into `div [||] []`.
- Removes `@@@react.dom` attribute as it's not needed anymore
- Updates ppx mapper to not be contextual
- Removes `dom_tags` check in ppx, as there's no need to validate lowercase elements at syntactic pass anymore

## TODO
- [x] Fix dangerouslySetInnerHtml
- [x] Re-enable ppx_errors* tests
- [x] Remove all appearances of `@@@react.dom`
- [x] Fix example components
- [x] Rename all functions with `_` at the end (see [comment](https://github.com/ml-in-barcelona/jsoo-react/pull/128#issuecomment-1024934071) below)
- [x] Think about custom elements

## Out of scope

- Update dsl functions so that each element can only receive allowed props (effectively merge ppx `Html` module into `Dsl.Html`).


jchavarri:
> It seemed to me that fully qualifying all the DSL calls (e.g. React.Dom.Dsl.Html.div) would make the transition seamless. But now that I've thought it through a bit more, I realize this will only work for known HTML elements. It wouldn't work for SVG or anything else, unless we have some convention for "escaping" the HTML scope.

@glennsl bringing [your comment](https://github.com/ml-in-barcelona/jsoo-react/pull/119#issuecomment-1024899669) here :) so we can continue the discussion in context of this PR. I think we have to experiment with this approach a bit. This might also guide us on the way on how to split the `React.Dom.Dsl` functions into separate modules.

For a first version, in this PR I am just doing very dumb transformation, and assuming user is opening the modules that are needed for the functions to type check. This might lead to issues, but I think it's easy to understand what the ppx is doing, which is also an advantage.

The only "magic" that I had to do for now is prefix the prop function with `Prop` to allow components to use props like `html`:

https://github.com/ml-in-barcelona/jsoo-react/blob/a677144d50b42b44542e025bf4854f791e84fac6/ppx/ppx.ml#L984

Also, a renaming of `ref` prop:

https://github.com/ml-in-barcelona/jsoo-react/blob/a677144d50b42b44542e025bf4854f791e84fac6/ppx/ppx.ml#L979

I am not sure how harmful this will be yet 😬

lib/dom_html.ml
```diff
@@ -461,8 +461,8 @@ module Prop = struct
 
   (* react-specific *)
 
-  let dangerouslySetInnerHTML ~(__html : string) =
-    any "dangerouslySetInnerHTML" (Dom.makeInnerHtml ~__html)
+  let dangerouslySetInnerHTML (safe_html : Dom.DangerouslySetInnerHTML.t) =
+    any "dangerouslySetInnerHTML" safe_html
```
glennsl, line 465: What's the reason for this change? Just uniformity?

glennsl: Uniformity for the sake of making the PPX simpler I assume?

glennsl:
> Also, a renaming of `ref` prop

Ah, yeah that's a bit unfortunate. Technically it's still possible to name the prop `ref`, but that would shadow `Stdlib.ref` so that you have to qualify that every time you use it. Not including `Prop` in `Html`, and only opening `Prop` locally is also a way we can go. But if we want `Prop` to be opened non-locally I think we have to name it `ref_` to avoid the shadowing.

There's also a few others that need renaming, btw. `method_`, ´open_` and `type_` in `Html`. And `begin_`, `end_`, `in_` and `to_` in `Svg`.

lib/dom_html.ml
jchavarri: Yes, exactly.

That, and also, I thought that it would makes sense to force consumers to call `React.Dom.makeHtml` explicitly, to emulate what React.js does with the `{__html}` object: https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html. The idea as I understand from the link above is to avoid situations where some string is passed directly.

jchavarri:
> There's also a few others that need renaming, btw. `method_`, `open_` and `type_` in `Html`. And `begin_`, `end_`, `in_` and `to_` in `Svg`.

@glennsl as i was looking into this, I thought that based on past experiences, we should try to minimize the amount of magic that happens in the ppx. So, I think it's ok if we just make the transformation always the same, and Reason jsx needs to use the real function name, e.g. `ref_`: https://github.com/ml-in-barcelona/jsoo-react/pull/128/commits/a4abfa6fea8b2f4691272acd79f16bc7155f83d7

What do you think?

Another thing, I saw `Dom_html.float` has no underscore but there's a function in OCaml stdlib with same name. What's the rule of thumb to append `_` to the `Dom_html` function name?

lib/dom_html.ml
glennsl: From an OCaml perspective, I think this is actually quite misleading, because you'd expect a construct like this to actually do something, like maintain some invariant. And I'd rather use something that actually does, and communicates that more clearly, like `SafeString.t`.

The problem is that the concept of "safe" here is extremely vague. It's a string containing only the "safe" set of HTML tags, where that set depends entirely on what's expected in a given context. It's difficult enough to create a HTML sanitizer, even more so to create a well-typed one that can sanitize selectively. We could start by making the type though, something like ``[ `a | `em | `strong | `p | `br ] Sanitized.t``, and for now just have a `val make_unchecked : string -> _ t` to create it, effectively asserting that it's already been sanitized.

glennsl:
> What do you think?

I think that's probably fine, as long as we're consistent.

> Another thing, I saw `Dom_html.float` has no underscore but there's a function in OCaml stdlib with same name. What's the rule of thumb to append `_` to the `Dom_html` function name?

The one's I've already suffixed have all been done just to fix compiler errors. Either because they're reserved keywords, or because I've actually encountered a problem with shadowing. The first one is simple enough, as the compiler will flat out refuse to compile if you use reserved keywords as identifiers. And the second one you could generalize to "anything that shadows the definitions in `Stdlib`". Then again, `Stdlib` is full of archaic historical artifacts and you could argue that you should use `float_of_int` or `Float.of_int` instead. But `float` isn't actually deprecated either.

Unless there are other counterarguments, I think we should go with the simplest rule: Avoid shadowing anything defined in `Stdlib`.

lib/dom_html.ml
jchavarri: I like this idea. I implemented it in https://github.com/ml-in-barcelona/jsoo-react/pull/128/commits/7767fc1a62f0530ea2e62ca50ff43672f50e522a. For now I didn't add any params to `SafeString.t` as I think it'd be more confusing while there are no checks being done. I added versions of `make_unchecked` with camel case and snake case, as I'd like to have all APIs supporting both style in the future 😅 
lmk what you think.

ppx/test/input_errors_01_element_attribute.ml
```diff
@@ -1,3 +0,0 @@
-[@@@react.dom]
-
-let%component make () = div ~lola:1 ()
```
jchavarri, line 3: This test does not make any sense anymore, as the ppx does not check that the prop name is valid.

lib/dom_dsl_core.ml
```diff
@@ -9,7 +9,7 @@ module Prop = struct
 
   let int key (value : int) = any key value
 
-  let float key (value : float) = any key value
+  let float_ key (value : float) = any key value
```
jchavarri, line 12: I updated this value of `float_` to not collide. But the other one in `Core` is left as is, as in that case collisions should not happen:

https://github.com/jchavarri/jsoo-react/blob/7767fc1a62f0530ea2e62ca50ff43672f50e522a/lib/dom_dsl_core.ml#L45

ppx/ppx.ml
```diff
@@ -1376,47 +1210,33 @@ let jsxMapper () =
               module name." )
   in
   object (self)
-    inherit [Context.t] Ast_traverse.map_with_context as super
```
jchavarri, line 1379: This should make the ppx faster and less heavy on memory 🎉

test/test_reason.re
```diff
@@ -1093,6 +1150,126 @@ let testWithId = () => {
...
+    );
+    assert_equal(c##.innerHTML, Js.string("<div class=\"foo\"></div>"));
+  });
+
+let testPropCustomString = () =>
+  withContainer(c => {
+    module Prop = {
+      let foo = Prop.string("foo");
+    };
+    act(() => React.Dom.render(<div foo="bar" />, Dom_html.element(c)));
```
jchavarri, line 1177: It is possible with some overrides to create invalid elements, but I don't think it's a big issue.

test/test_reason.re
```diff
@@ -1093,6 +1150,126 @@ let testWithId = () => {
...
+  withContainer(c => {
+    let cool_element = h("cool-element");
+    let chill = h("chill");
+    module Prop = {
+      let coolness = Prop.string("coolness");
+      let data_out = Prop.bool("data-out");
+    };
+    act(() =>
+      React.Dom.render(
+        <cool_element coolness="max"> <chill data_out=true /> </cool_element>,
```
jchavarri, line 1232: Example of custom elements + jsx.

jchavarri (commented):
@glennsl  @davesnx  This is ready to review. The remaining step would be to migrate the `Html` module in ppx so that the functions in the DSL become "tag dependent". This can be done in a future PR.

lib/dom.mli
```diff
@@ -1503,9 +1503,16 @@ module Style : sig
...
+module SafeString : sig
   type t
-end
 
-val makeInnerHtml : __html:string -> DangerouslySetInnerHTML.t
-  [@@js.builder] [@@js.verbatim_names]
+  val make_unchecked : string -> t
+    [@@js.custom
+      val make_react_html_object : __html:string -> t
+        [@@js.builder] [@@js.verbatim_names]
```
glennsl, line 1512: I think `t` ought to be a `private string`, and have `dangerouslySetInnerHTML` create this object instead. That would allow this type to be more generally useful. It's also more idiomatic OCaml.

ppx/test/input_ocaml.ml
```diff
@@ -1,24 +1,18 @@
-[@@@react.dom]
+let%component make ?(name = "") = div [||] [React.string ("Hello " ^ name)]
 
-let%component make ?(name = "") =
-  div ~children:[React.string ("Hello " ^ name)] ()
+let element = div [|id "foo"|] []
 
-let element = div ~id:"foo" ~children:[] ()
+let element = a [|id "foo"; href "https://www.hello.com"|] [React.string "hello"]
```
glennsl, line 5: I would think these can be removed, as there's no syntax transformation going on here anymore.

ppx/test/input_ocaml.ml
```diff
@@ -1,24 +1,18 @@
...
+let%component make ~children:(first, second) () = div [||] [first; second]
 
-let%component make ~children:(first, second) () =
-  div ~children:[first; second] ()
+let%component make ?(name = "") = div [||] [name]
 
-let%component make ?(name = "") = div ~children:[name] ()
-
-let%component make () = a ?href:(Some "https://opam.ocaml.org") ()
+let%component make () = a [|href "https://opam.ocaml.org"|] []
```
glennsl, line 15: This one too.

test/test_reason.re
glennsl: Oh wow, React will already do that? I did not know!

glennsl (commented):
> I added versions of make_unchecked with camel case and snake case, as I'd like to have all APIs supporting both style in the future

I'm questioning this a little bit. Is there a reason to support camelCase at all, apart from compatibility with react and rescript-react? Everything else in jsoo-land is snake_cased right? I would certainly strongly prefer to have the same casing used throughout the whole ecosystem, so I don't have to remember which case to use for which functions. I also fear the documentation is going to be very messy if we're going to have two versions of everything.

Would it be possible to make it snake_case only by default, but if necessary provide a compatibility layer of some kind to get camelCased versions? That would both signal a preference and make the documentation less messy by separation.

lib/dom.mli
jchavarri: Having `dangerouslySetInnerHTML` create the `__html` object makes sense.

I moved some code in https://github.com/ml-in-barcelona/jsoo-react/pull/128/commits/d3924adb125a8a7fadb8c2535486127c9b05c1fa, let me know if this is somehow what you had in mind. As we have now `Dom` and `Dom_html` and it's unclear to me what should go where 😄

test/test_reason.re
jchavarri: React will create DOM elements with whatever tag is passed to `createElement`. Here we pass `cool-element` through the `h` function:

https://github.com/ml-in-barcelona/jsoo-react/blob/d3924adb125a8a7fadb8c2535486127c9b05c1fa/test/test_reason.re#L1224

Because we have not defined these elements Right now we get a warning when running the tests:

```
Warning: The tag <chill> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.
    in chill
    in cool-element
```

I am not sure I get it, which is the part you did not know about? :)

jchavarri:
> I'm questioning this a little bit. Is there a reason to support camelCase at all, apart from compatibility with react and rescript-react? Everything else in jsoo-land is snake_cased right? 

Yeah, I agree that keeping both will be a mess in terms of documentation. I removed the camel case in https://github.com/ml-in-barcelona/jsoo-react/pull/128/commits/d3924adb125a8a7fadb8c2535486127c9b05c1fa.

Most probably the Reason "compat layer" should be isolated on its own package, where all redefinitions using uppercase could be performed. This kind of relates with st else I've been thinking, which is splitting the ppx in two, as all the `@JSX` attrs processing is unnecessary for OCaml syntax users, so they should not pay the costs and complexity. So maybe makes sense to package all the Reason stuff together in some `jsoo-react-reason` package 🤔 

It kind of bothers me that all `Core` functions use camelCase, like `useState` etc. We should probably revisit them as well.

lib/dom.mli
glennsl: LGTM. Thanks!

test/test_reason.re
glennsl: Oh, I thought React was the one translating `cool_element` to `cool-element`, but it's the PPX. Yeah that's a bit of a shame with the warning, it ought to accept standard custom element names I think.

glennsl:
Yeah I think having a separate PPX and `jsoo-react-reason` package is a great idea. Makes everything much cleaner. As long as they can be mixed, as I imagine some may want to try out both or migrate gradually from one to the other.

> It kind of bothers me that all Core functions use camelCase, like useState etc. 

Me too. If given permission I can clean those up in the near future :grin:

test/test_reason.re
jchavarri: I think it'd work if we had previously registered `cool-element` and `chill` as custom elements using `window.customElements.define`. I tried, but unfortunately jsoo won't parse `class Foo extends HTMLElement {...}`... so I guess we will have to live with this warning for now.

glennsl: Hmm, I just got this to work in #129. Didn't try Reason syntax though. I'm not sure how React would check the custom element rehister though, or if that's even possible.

jchavarri: > Didn't try Reason syntax though. I'm not sure how React would check the custom element register though, or if that's even possible.

It does some hacks to check if the element has been registered apparently: https://github.com/facebook/react/blob/24dd07bd269590ee5024b7f0f1906887d256ea86/packages/react-dom/src/client/ReactDOMComponent.js#L447-L461

I don't understand the `if` condition though, as it uses `if (!isCustomComponentTag && ...`, but I would expect `if (isCustomComponentTag && ...`, without negation 🤔 

> Didn't try Reason syntax though.

Nothing should change, right? The Reason syntax conversion of JSX elements is now super straight forward.

jchavarri:
> Me too. If given permission I can clean those up in the near future

Yes! Sounds great, this would be super appreciated 🙏

test/test_reason.re
glennsl: > I don't understand the `if` condition though, as it uses `if (!isCustomComponentTag && ...`, but I would expect `if (isCustomComponentTag && ...`, without negation thinking

I don't know what `isCustomComponent` means, but I think it's the second clause, `Object.prototype.toString.call(domElement) === '[object HTMLUnknownElement]'`. The trick seems to be to check after instantiating the dom element, whether it instantiated a known or unknown element.

> Nothing should change, right? The Reason syntax conversion of JSX elements is now super straight forward.

That's true! I guess it's alright then :)

davesnx (approved):

example/src/Code.re
```diff
@@ -1,4 +1,6 @@
 open Js_of_ocaml;
+open React.Dom.Dsl;
```
davesnx, line 2: I missed that on another PR, but any thought on having React.DOM.DSL?

davesnx: @jchavarri @glennsl

glennsl: I have no particular opinion :neutral_face:

davesnx: True, not super strong on those things, just read Ds1 in my head and broke it

jchavarri: Another advantage of current api is that if we decide to move it to a file `dom.mli` or `dsl.mli` in the future, the api / module names would not need to change. To avoid changes in uppercase, the files would need to be called `DOM.mli` or `DSL.mli`.

PR #129, opened 2022-01-30
add Webcomponent example

glennsl:
This might be a bit offensive to the Barcelonians, but it's based on the example here: https://github.com/glennsl/leafy-example

Maybe we could find a simple Geojson polygon of Barcelona or Catalonia to use here instead?


jchavarri (approved):
This is so cool!

> This might be a bit offensive to the Barcelonians
> Maybe we could find a simple Geojson polygon of Barcelona or Catalonia to use here instead?

haha, I think Oslo map is great :) I am convinced @davesnx agrees!

davesnx (approved):

jchavarri:
updated example in https://ml-in-barcelona.github.io/jsoo-react/.

PR #138, opened 2022-02-06
Convert API to snake case

glennsl:
While we're at it making breaking changes, I'd like to follow up with some reorganization and renaming to make the API more ergonomic, such as giving `useEffect0` etc. more meaningful names, dropping unnecessary repetition and verbosity, such as in `React.Dom.create_dom_element_variadic`, using more idiomatic OCaml constructs such as labeled arguments instead of putting `_with_X` in function names, and generally try to clean things up a bit. But I'll do that in a later PR once this has been approved and merged.

### DIFF
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -8,7 +8,7 @@
 (*
    The transform:
    transform `[@JSX] div(~props1=a, ~props2=b, ~children=[foo, bar], ())` into
-   `React.Dom.createDOMElementVariadic("div", React.Dom.domProps(~props1=1, ~props2=b), [foo, bar])`.
+   `React.Dom.create_dom_element_variadic("div", React.Dom.domProps(~props1=1, ~props2=b), [foo, bar])`.
    transform the upper-cased case
    `[@JSX] Foo.createElement(~key=a, ~ref=b, ~foo=bar, ~children=[], ())` into
    `React.create_element(Foo.make, Foo.makeProps(~key=a, ~ref=b, ~foo=bar, ()))`
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -674,7 +674,7 @@ let make_js_comp ~loc ~fn_name ~forward_ref ~has_unit ~named_arg_list
 
 (* Builds the intermediate function with labelled arguments that will call make_props.
    [body] is the the component implementation as originally written in source,
-   but without any wrappers like React.memo or forwar_ref *)
+   but without any wrappers like React.memo or forward_ref *)
 let make_ml_comp ~loc ~fn_name ~body rest =
   Exp.mk ~loc
     (Pexp_let (Nonrecursive, [Vb.mk (Pat.var {loc; txt= fn_name}) body], rest))
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -11,10 +11,10 @@
    `React.Dom.createDOMElementVariadic("div", React.Dom.domProps(~props1=1, ~props2=b), [foo, bar])`.
    transform the upper-cased case
    `[@JSX] Foo.createElement(~key=a, ~ref=b, ~foo=bar, ~children=[], ())` into
-   `React.createElement(Foo.make, Foo.makeProps(~key=a, ~ref=b, ~foo=bar, ()))`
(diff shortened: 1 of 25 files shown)


ppx/ppx.ml
```diff
@@ -674,7 +674,7 @@ let make_js_comp ~loc ~fn_name ~forward_ref ~has_unit ~named_arg_list
 
 (* Builds the intermediate function with labelled arguments that will call make_props.
    [body] is the the component implementation as originally written in source,
-   but without any wrappers like React.memo or forwardRef *)
+   but without any wrappers like React.memo or forwar_ref *)
```
jchavarri, line 677:
```suggestion
   but without any wrappers like React.memo or forward_ref *)
```

test/test_reason.re
```diff
@@ -28,7 +28,7 @@ let withContainer = f => {
   let container = Dom_html.createDiv(doc);
   Dom.appendChild(doc##.body, container);
   let result = f(container);
-  ignore(React.Dom.unmountComponentAtNode(container));
+  ignore(React.Dom.unmount_component_at_node(container));
```
jchavarri, line 31: lol this test file is "disintegrating" over time, little by little 😄

jchavarri (approved):
This looks really good! Thanks a lot for taking the time to do all this work.

The changes seem quite mechanical, so I have nothing to add at specific level except one tiny typo.

On a more generic level, my main concern is that this is effectively removing ergonomic support for Reason syntax. I guess in the future we could have a separate lib that just aliases the OCaml names? I am not sure how to do this without it becoming a maintenance burden unfortunately (same thing has happened to me in the past with docs, examples, etc). Maybe we could automate it with some script, or ppx. Not sure. Tablecloth tried to do this but you can see how much effort it takes, and challenges like some flavour being documented while the other [is not](https://github.com/darklang/tablecloth/blob/349232b6214e60f582d629e69e2d51fed9d9f37e/rescript/src/TableclothArray.mli#L140)).

If we don't find an easy way to add Reason camel case names, I would most probably remove the JSX part away from the ppx after this PR is merged, and move it to a separate ppx, or even package or repo, so there is less area for confusion (and frustration).

I leave the PR open for 1 more day or so, to leave time for others to comment :)

glennsl:
> On a more generic level, my main concern is that this is effectively removing ergonomic support for Reason syntax.

This seems to conflate syntax and ergonomics with convention and familiarity. I don't think there's anything about the Reason syntax that actually makes snake_cased identifiers less _ergonomic_. I suppose it is in Reason's _"spirit"_ to use the more familiar camelCased identifiers, but operating in the OCaml ecosystem with this convention has always been quite awkward and unergonomic, I think, because most of the ecosystem does not follow that convention and so you have to remember which APIs use which convention. It was only in the BuckleScript ecosystem that this worked well, because that ecosystem was built from the ground up on those conventions.

Not that familiarity can't also be a goal, of course, but even if it's feasible to maintain two different naming conventions for jsoo-react, how feasible is it really to do so for the rest of the ecosystem? One can perhaps hope that the relative popularity of Rust has made the combination of snake_case and Algol-style syntax more familiar, so that it wouldn't be that much of a hurdle.

jchavarri:
> I suppose it is in Reason's "spirit" to use the more familiar camelCased identifiers, but operating in the OCaml ecosystem with this convention has always been quite awkward and unergonomic,

Yeah what I meant is that historically Reason has "inherited" JavaScript camel case in the APIs. Even Js_of_ocaml has followed similar approach as well ([example](https://github.com/ocsigen/js_of_ocaml/blob/882f26c5826ba7839fdc9270d4dfb987f17a7396/lib/js_of_ocaml/dom_html.mli)).

But more recent packages like [Brr](https://erratique.ch/software/brr/doc/Brr_canvas/Gl/index.html) moves in the same direction as this PR: using OCaml snake case names everywhere, which I think makes more sense and removes confusion in the long term.

If there are users of jsoo-react that want to keep using Reason syntax and would like some "flavour" of the jsoo-react library using Reason/camel case names I think we can discuss how to implement it in a (hopefully automated) way to reduce maintenance costs.

jchavarri:
Thanks again for your work on this, @glennsl !

PR #139, opened 2022-02-08
Docs: experiment with docsify

jchavarri:
This is a small experiment for what could be the first version of docs (not counting existing `example` folder, which served its purpose, but I think can be retired now 😅 ).

docs

This experiment is using [docsify](https://docsify.js.org/) with some updates. I really liked docsify for a few reasons:

- It is very minimal and clean
- It is well integrated with markdown
- It support full text search out of the box
- And multi-lang, which could be useful
- Can be read from mobile
- No build step, just an html file (can be served with `python -m SimpleHTTPServer`, `docsify-cli` or any other server)

Some more things:

## Embed `.ml` files

Through a [tiny change](https://github.com/docsifyjs/docsify/compare/develop...jchavarri:ocaml) in docsify, there is the possibility to embed `ml` files partially, only the part between comments with `(* [demo] *)` in them.

This allows to have the doc snippets compiled with dune (including ppx, etc) very easily, apply ocamlformat to them transparently too, and then use them in the docs directly, see [docs/snippets/basic.ml](https://github.com/ml-in-barcelona/jsoo-react/compare/docs-v1?expand=1#diff-9aa59d625961628f96a6f164e2ed525e404d76208344285f70563644d6009661) and [how it renders](https://ml-in-barcelona.github.io/jsoo-react/docs/#/first_components). 

## Embed html files from odoc

It is possible to embed html in docsify as well. This allows to produce some html with odoc, and then just copy the content to the `docs` folder to be shown with the rest of the docs, see [example](https://ml-in-barcelona.github.io/jsoo-react/docs/#/api.html).

One question here is if we will be able to put all the docs at the top level component (`React`) because otherwise I don't think the embedding will work, as it'll be very hard to maintain (one has to reproduce the hierarchy of odoc manually in docsify). This is probably the part that makes me hesitate more.

## Docs initial structure

I added as well some initial structure for the docs (more details about subsections in link below):
- Introduction
- Main concepts
- Guides
- API

The example to see how all the above works can be seen in https://ml-in-barcelona.github.io/jsoo-react/docs/. 

Let me know what you think :) any feedback welcome, about structure, choice of libs, embedding etc.


docs/first_components.md
```diff
@@ -0,0 +1,7 @@
+# Your first components
+
+## A simple component
+
+Jsoo-react components are functions that take input data and returns what to display. The library provides a DSL that allows to create elements like `div`, or `p` with just plain OCaml functions. The uppercase component props are passed as labelled arguments, while lowercase components take an array of attributes.
+
+[filename](snippets/basic.ml ':include :type=code :fragment=demo')
```
davesnx, line 7: :clever:

docs/index.html
```diff
@@ -0,0 +1,41 @@
...
+        placeholder: 'Type to search',
+        noData: 'No results',
+        // Headline depth, 1 - 6
+        depth: 6,
+        hideOtherSidebarContent: false, // whether or not to hide other sidebar content
+     }
+    }
+  </script>
+  <!-- Docsify v4 patched -->
+  <script src="//cdn.jsdelivr.net/npm/@jchavarri/docsify@4.12.2-alpha.1"></script>
```
davesnx, line 33: 👍🏼

davesnx (approved):

PR #141, opened 2022-02-13
Experiment: more ergonomic/idiomatic API

glennsl:
This aims to adapt the API to be more ergonomic and idiomatic OCaml. The purpose being to get a thorough understanding of what an API that is internally consistent with the OCaml ecosystem would look like, and what a developer coming from an OCaml background, rather than a React background, would expect. Based on this, and with the existing React-like API as a counter-point, we'll hopefully be able to come up with a synthesis that will be familiar to both React and OCaml developers, and provide a consistent experience when working with the rest of the OCaml ecosystem.

### DIFF
--- a/lib/core.mli
+++ b/lib/core.mli
@@ -329,26 +329,11 @@ val use_effect_on_change7 :
       (if before_render then _use_layout_effect7 else _use_effect7)
         Imports.react f values]
 
-val use_callback : ('input, 'output) callback -> ('input, 'output) callback
-  [@@js.custom
-    val use_callback_internal :
-         Imports.react
-      -> ('input, 'output) callback
-      -> 'a array option_undefined
-         (* Important: we don't want to use an arrow type to represent returning callback (i.e. (Ojs.t -> Ojs.t)) as the callback
-            would get wrapped inside caml_js_wrap_callback_strict in the resulting code *)
-      -> ('input, 'output) callback
-      [@@js.call "useCallback"]
-
-    let use_callback cb = use_callback_internal Imports.react cb None]
-
-val use_callback0 : ('input, 'output) callback -> ('input, 'output) callback
-  [@@js.custom
-    let use_callback0 f = use_callback_internal Imports.react f (Some [||])]
-
 val use_callback1 :
   ('input, 'output) callback -> 'a array -> ('input, 'output) callback
   [@@js.custom
+    (* Important: we don't want to use an arrow type to represent returning callback (i.e. (Ojs.t -> Ojs.t)) as the callback
+       would get wrapped inside caml_js_wrap_callback_strict in the resulting code *)
     val use_callback1_internal :
          Imports.react
       -> ('input, 'output) callback
--- a/lib/core.mli
+++ b/lib/core.mli
@@ -687,25 +687,17 @@ module Fragment : sig
           (make_props ?key ()) children]
 end
 
-val memo : 'props component -> 'props component
+val memo :
+  ?compare:('props -> 'props -> bool) -> 'props component -> 'props component
   [@@js.custom
-    val memo_internal : Imports.react -> 'props component -> 'props component
-      [@@js.call "memo"]
-
-    let memo component = memo_internal Imports.react component]
-
-val memo_custom_compare_props :
-  'props component -> ('props -> 'props -> bool) -> 'props component
-  [@@js.custom
-    val memo_custom_compare_props_internal :
+    val memo_internal :
          Imports.react
       -> 'props component
-      -> ('props -> 'props -> bool)
+      -> ('props -> 'props -> bool) option_undefined
       -> 'props component
       [@@js.call "memo"]
 
-    let memo_custom_compare_props component compare =
-      memo_custom_compare_props_internal Imports.react component compare]
+    let memo ?compare component = memo_internal Imports.react component compare]
 
 val set_display_name : 'props component -> string -> unit
   [@@js.set "displayName"]
--- a/lib/core.mli
+++ b/lib/core.mli
@@ -428,17 +428,6 @@ val use_callback7 :
     let use_callback7 callback watchlist =
       use_callback7_internal Imports.react callback watchlist]
 
-val use_memo : (unit -> 'value) -> 'value
-  [@@js.custom
-    val use_memo_internal :
-      Imports.react -> (unit -> 'value) -> 'any array option_undefined -> 'value
-      [@@js.call "useMemo"]
-
-    let use_memo f = use_memo_internal Imports.react f None]
-
-val use_memo0 : (unit -> 'value) -> 'value
-  [@@js.custom let use_memo0 f = use_memo_internal Imports.react f (Some [||])]
-
 val use_memo1 : (unit -> 'value) -> 'a array -> 'value
   [@@js.custom
     val use_memo1_internal :
--- a/lib/core.mli
+++ b/lib/core.mli
@@ -667,14 +667,13 @@ module Children : sig
 
       let map element mapper = map_internal children_internal element mapper]
 
-  val map_with_index : element list -> (element -> int -> element) -> element
+  val mapi : element list -> (element -> int -> element) -> element
     [@@js.custom
-      val map_with_index_internal :
+      val mapi_internal :
         Ojs.t -> element list -> (element -> int -> element) -> element
         [@@js.call "map"]
 
-      let map_with_index element mapper =
-        map_with_index_internal children_internal element mapper]
+      let mapi element mapper = mapi_internal children_internal element mapper]
 
   val for_each : element list -> (element -> unit) -> unit
     [@@js.custom
--- a/lib/core.mli
+++ b/lib/core.mli
@@ -104,245 +104,41 @@ val clone_element : element -> 'props -> element
     let clone_element element props =
       clone_element_internal Imports.react element props]
 
-val use_effect : (unit -> (unit -> unit) option_undefined) -> unit
-  [@@js.custom
-    val use_effect_internal :
-      Imports.react -> (unit -> (unit -> unit) option_undefined) -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect callback = use_effect_internal Imports.react callback]
-
-val use_effect0 : (unit -> (unit -> unit) option_undefined) -> unit
-  [@@js.custom
-    val use_effect0_internal :
-         Imports.react
-      -> (unit -> (unit -> unit) option_undefined)
-      -> Ojs.t array
-      -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect0 callback = use_effect0_internal Imports.react callback [||]]
-
-val use_effect1 : (unit -> (unit -> unit) option_undefined) -> 'a array -> unit
-  [@@js.custom
-    val use_effect1_internal :
-         Imports.react
-      -> (unit -> (unit -> unit) option_undefined)
-      -> 'a array
-      -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect1 callback watchlist =
-      use_effect1_internal Imports.react callback watchlist]
-
-val use_effect2 : (unit -> (unit -> unit) option_undefined) -> 'a * 'b -> unit
-  [@@js.custom
-    val use_effect2_internal :
-         Imports.react
-      -> (unit -> (unit -> unit) option_undefined)
-      -> 'a * 'b
-      -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect2 callback watchlist =
-      use_effect2_internal Imports.react callback watchlist]
-
-val use_effect3 :
-  (unit -> (unit -> unit) option_undefined) -> 'a * 'b * 'c -> unit
-  [@@js.custom
-    val use_effect3_internal :
-         Imports.react
-      -> (unit -> (unit -> unit) option_undefined)
-      -> 'a * 'b * 'c
-      -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect3 callback watchlist =
-      use_effect3_internal Imports.react callback watchlist]
-
-val use_effect4 :
-  (unit -> (unit -> unit) option_undefined) -> 'a * 'b * 'c * 'd -> unit
-  [@@js.custom
-    val use_effect4_internal :
-         Imports.react
-      -> (unit -> (unit -> unit) option_undefined)
-      -> 'a * 'b * 'c * 'd
-      -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect4 callback watchlist =
-      use_effect4_internal Imports.react callback watchlist]
-
-val use_effect5 :
-  (unit -> (unit -> unit) option_undefined) -> 'a * 'b * 'c * 'd * 'e -> unit
-  [@@js.custom
-    val use_effect5_internal :
-         Imports.react
-      -> (unit -> (unit -> unit) option_undefined)
-      -> 'a * 'b * 'c * 'd * 'e
-      -> unit
-      [@@js.call "useEffect"]
-
-    let use_effect5 callback watchlist =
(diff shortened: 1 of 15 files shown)


jchavarri:
Thanks for this @glennsl, I haven't been able to review it yet, but checked the issues to unblock you.

> Also, any ideas on troubleshooting hanging test issues?

I just commented all effects related tests and then uncomment progressively, until I found the culprit. I noticed that uncommenting this call to `use_effect_on_change1` would trigger the problem again:

https://github.com/glennsl/jsoo-react/blob/6ffbb3bd4406b398d2cf2f2e9b537bacc6318b41/test/test_reason.re#L157

After checking the code, I saw that after the changes in this branch, the dependency for `use_effect_on_change1` is now automatically wrapped from the library code:

https://github.com/glennsl/jsoo-react/blob/6ffbb3bd4406b398d2cf2f2e9b537bacc6318b41/lib/core.mli#L181

while this was not the case before:

https://github.com/glennsl/jsoo-react/blob/5b81ab32cd4fc95277da8bb2ab97842e2b7656dd/lib/core.mli#L131

Unwrapping the dependency in [the test](https://github.com/glennsl/jsoo-react/blob/6ffbb3bd4406b398d2cf2f2e9b537bacc6318b41/lib/core.mli#L181) fixes the issue:

```ocaml
React.use_effect_on_change1(
  () => {
    setCountStr(((count, str)) => (count + 1, f(str)));
    None;
  },
  f,
);
```

glennsl:
Ah, thanks @jchavarri. Good catch!

I think the root problem of the infinite loop in particular is a test design problem I've also come across with other tests, that they use `use_state` to track test state, which itself also affects the test because it triggers re-rendering. And because a new array is created for the watched variable every time, it will always be invalidated, call the effect and trigger another re-rendering etc., ad infinitum. I've converted a few tests to use an ordinary `ref` instead to avoid this problem, and probably should go over all the tests to do the same at some point.

jchavarri:
> And because a new array is created for the watched variable every time, it will always be invalidated, call the effect and trigger another re-rendering etc., ad infinitum.

This makes sense. It's a bit unfortunate that we have to stop using `use_state` in tests due to this issue.

glennsl:
We don't necessarily have to stop using `use_state`, I just don't think it's a good idea to have the test setup affect the tests because it can lead to issues like this, and even obscure actual test failures.

Why do you think it's unfortunate to stop using it?

jchavarri:
I was just thinking that the more `use_state` is used in tests, the better (more edge cases can be caught). But it is probably not necessary, as long as we have some specific tests for it, then it makes sense to not use it and simplify other tests like effects etc. 👍

lib/core.mli
```diff
@@ -99,233 +99,230 @@ val clone_element : element -> 'props -> element
     let clone_element element props =
       clone_element_internal Imports.react element props]
 
-val use_effect : (unit -> (unit -> unit) option_undefined) -> unit
+val use_effect_always :
+  ?before_render:bool -> (unit -> (unit -> unit) option_undefined) -> unit
```
jchavarri, line 103: `before_render` adds cognitive load as there's nothing like this in either React or rescript-react, and also leads to more verbose code. What are the reasons why you think `use_layout_effect*` functions should be replaced with it?

lib/core.mli
```diff
@@ -99,233 +99,230 @@ val clone_element : element -> 'props -> element
...
+      -> 'a array
       -> unit
-      [@@js.call "useEffect"]
+      [@@js.call "useLayoutEffect"]
 
-    let use_effect5 callback watchlist =
-      use_effect5_internal Imports.react callback watchlist]
+    let use_effect_on_change1 ?(before_render = false) f a =
+      (if before_render then _use_layout_effect else _use_effect)
+        Imports.react f [|a|]]
```
jchavarri, line 181: This is 100%opinion, but I think this change is going to lead to a lot of confusion as diverges from rescript-react, as was seen with the tests endless rendering. Why not leave the array wrapping outside the function?

lib/core.mli
```diff
@@ -536,56 +533,35 @@ val use_state : (unit -> 'state) -> 'state * (('state -> 'state) -> unit)
...
-          [| Ojs.fun_to_js 2 (fun (state : Ojs.t) (action : Ojs.t) ->
-                 Obj.magic (reducer (Obj.magic state) (Obj.magic action)) )
-           ; Obj.magic init |]
-      in
-      (Obj.magic (Ojs.array_get result 0), Obj.magic (Ojs.array_get result 1))
-
-    let use_reducer reducer initial =
-      use_reducer_internal Imports.react reducer initial]
-
-val use_reducer_with_map_state :
```
jchavarri, line 561: What happened with this function? It is useful to do lazy initialization https://reactjs.org/docs/hooks-reference.html#lazy-initialization

lib/core.mli
```diff
@@ -536,56 +533,35 @@ val use_state : (unit -> 'state) -> 'state * (('state -> 'state) -> unit)
...
-     ('state -> 'action -> 'state)
-  -> 'initialState
-  -> ('initialState -> 'state)
+     reducer:('state -> 'action -> 'state)
```
jchavarri, line 536: Maybe we could put `reducer` last, and remove the label. Otherwise it's a bit strange to do `use_reducer ~reducer` 🤔

test/test_reason.re
```diff
@@ -146,103 +146,6 @@ let testContext = () => {
   });
 };
 
-let testUseEffect = () => {
```
jchavarri, line 149: Deleted because already available in OCaml file, right?

jchavarri (commented):
This is looking great, I left comments on parts where things are not so clear to me.

lib/core.mli
glennsl: `before_render` is very descriptive, signalling that this will trigger the effect before rendering, as opposed to after, and that this is the only effect the flag will have. `use_layout_effect` on the other hand sounds like a completely different kind of effect. What is a "layout effect"? What does it do, and how is it different from an ordinary effect? Given no further context, I think the latter has a much bigger cognitive load.

It's probably not fair to not consider other context, of course, such as existing familiarity with React. But for this iteration I'm trying to imagine how I'd design the API if I did so from scratch. Then, starting from that, try to take other considerations into account. So it would be helpful to know if your concerns are solely from the perspective of someone who already knows React, or if it's more general.

lib/core.mli
glennsl: It's leaking internal implementation details. There's no reason the user should need to know that this uses an array internally, and have to wrap it manually. It's also inconsistent with all the other `use_effect` functions, which all take a tuple instead of an array. Because, of course, tuples are represented as arrays in JS. But there is no 1-tuple that maps to a single-element array, because that makes absolutely no sense from an OCaml perspective. Would you consider an API like this reasonable if implemented in pure OCaml? If not, why should we expect user's to find it reasonable just because it binds to JavaScript?

I think the reason `rescript-react` has chosen this design is solely because they obsess over zero-cost bindings, irregardless of cognitive cost. If you interact very closely with JavaScript and expect your users to deeply understand this model, it might make sense, but I'm not sure that transfers all that well to jsoo.

You do point out a real problem though, and it's a hard one to fix without continuing the broken model of `rescript-react`...

lib/core.mli
glennsl: `use_reducer` does only lazy initialization instead now. Do you think there's a reason it should need to do eager initialization?

lib/core.mli
glennsl: Good idea. I was thinking you'd usually use a separate `reducer` function, which would make it equally repetitive anyway, and using a labelled argument would also help guide user's to naming that function. That's a very minor point though.

test/test_reason.re
glennsl: Yep! Although in a slightly different form.

lib/core.mli
glennsl: We could also still keep `useLayoutEffect` around, or provide a compatibility layer that has all the familiar react naming. The reason I'm making this "ideal"  version of the API first is because it will be easier to decide on a strategy when we can see both perspectives clearly in front of us.

> All this is explained with details in https://reactjs.org/docs/hooks-reference.html#uselayouteffect and https://reactjs.org/docs/hooks-reference.html#timing-of-effects. And prob a 100 other places :)

Thanks. I do already know this though :) It was more of a rhetorical question, with the point being that this is very non-obvious just from the name "layout effect", and for a novice would therefore raise these kinds of unnecessary questions.

glennsl:
A few more adaptations made, but still a work in progress. The last commit unfortunately fails to compile with the ppx error:

```
Fatal error: exception (Invalid_argument
  "react.component calls can only be on function definitions or component wrappers (forward_ref, memo).")
```

And unfortunately I'm not able to understand the function that emits this error:

https://github.com/ml-in-barcelona/jsoo-react/blob/8805668dd8a1c64b55eecfdd1719a52c45b336f9/ppx/ppx.ml#L693-L717

Some help with this would be greatly appreciated :) (Unless this change is a really bad idea for some reason)

jchavarri:
@glennsl this part of the ppx is quite hairy 🐻 it is the part that dives into the expression that has the `react.component` attribute to see what are the labelled args, so we can gather them to create the makeprops function. For the majority of cases the labelled args can be retrieved from the "top" expression, but for cases like `React.memo` we have to keep "digging" into the parsed tree to get these labelled args (hence the function name `spelunk_for_fun_expr`. Btw most of this code comes from reason-react ppx, I barely touched it, maybe there are easier ways to handle this.

I pushed a commit that fixed ppx and tests in https://github.com/ml-in-barcelona/jsoo-react/commit/744f3697e78c96a216bb225cbc5aa49cb62347c6, feel free to pick whatever you need or improve it.

There is one thing that if you don't pick I will add in another PR though, which is the locations for the `react.component calls can only be on function definitions...` error.

glennsl:
Thanks @jchavarri! I was completely missing the second location that had to be changed, and therefore couldn't figure out why it failed no matter how I tweaked the first pattern. Also really appreciate the better error reporting!

This mechanism does seem very fragile. But I think I improved it a little bit, with the help of having the argument labelled, by not requiring it to be a syntactic function and allowing it to be placed both before and after the unlabeled argument.

glennsl:
I'm pretty sure I've gone all the API tweaks I wanted to make now. As explained in comments, and also now in the OP, in this first iteration I wanted to answer the question "what would this API look like if it was designed purely from an OCaml perspective?". And I think in a first round of review it would be good to take that perspective as well. Then, once we have a clear understanding of the two "extremes", their benefits and problems, we can go about trying to figure out how to make the API approachable from both directions. A few alternatives we can pursue to that end:

- Compromise the API design by trying to eliminate "problems" from both perspectives, sacrificing some of the benefits along the way.
- Keep the primary API idiomatic OCaml, but add a compatibility layer providing a more faithful and familiar React API.
- Provide specific documentation targeting React developers to explain the differences and rationale.

lib/core.mli
jchavarri: > Do you think there's a reason it should need to do eager initialization?

I can't think of any reasons.

> This is probably the change I'm most uncertain about, from an idealist perspective at least.

It is funny because this is one of the less disruptive changes imo 😄 I believe we have different use cases in mind.

jchavarri (approved):
@glennsl Thanks for your work on this!

This looks good, I really like the changes to `Ref` and `Context`, and the new APIs are way more natural for OCaml developers.

I still have doubts about `use_effect` changes, especially the automatic wrapping in array happening in `use_effect_on_change1`, but my concerns are mostly based on potential migrations of existing reason-react apps. As we have discussed in other occasions this idea of having a `jsoo-react-reason` compatibility library, maybe we can keep reason-react approach over there, and we remove "pressure" from the core API being worked on here.

I will wait until the weekend before merging the PR, to see if anyone else has thoughts or comments.

lib/core.mli
```diff
@@ -104,254 +104,236 @@ val clone_element : element -> 'props -> element
...
+      [@@js.call "useLayoutEffect"]
 
-    let use_effect1 callback watchlist =
-      use_effect1_internal Imports.react callback watchlist]
+    let use_effect_once ?(before_render = false) f =
+      (if before_render then _use_layout_effect else _use_effect)
+        Imports.react f [||]]
 
-val use_effect2 : (unit -> (unit -> unit) option_undefined) -> 'a * 'b -> unit
+val use_effect_on_change :
```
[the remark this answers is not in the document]
glennsl: I think I just wanted it for easy migration and just in case, but it's hard to think of an actual use case, yeah. I agree it can probably just be removed.

lib/core.mli
```diff
@@ -667,31 +617,30 @@ module Children : sig
 
       let map element mapper = map_internal children_internal element mapper]
 
-  val map_with_index : element list -> (element -> int -> element) -> element
+  val mapi : element list -> (element -> int -> element) -> element
     [@@js.custom
-      val map_with_index_internal :
+      val mapi_internal :
```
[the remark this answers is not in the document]
glennsl: I got some name conflicts when I used the same name for some of these internal functions, and so I did some experiments to try to figure out what gen_js_api would actually generate, and in what way they "leak". I can't remember exactly what I concluded, but I think these function names leak out into the file namespace. So I started to think about how to minimize the noise this creates, that prefixing them with `_` prevent them from showing up in completion menus, as they would when suffixed with `_internal`, for example. It's also easier to rename prefixed names (in vim). It's a bit of a half-baked experiment, but so far I can't see any downsides to prefixing with `_` rather than suffixing with `_internal` at least.

test/test_ml.ml
```diff
@@ -112,96 +112,55 @@ let testContext () =
...
-        React.use_callback1 (fun input -> input ^ " " ^ a ^ " and") [|a|]
-      in
-      React.use_effect1
+      let f = React.use_callback1 (fun input -> input ^ " " ^ a ^ " and") a in
+      React.use_effect_on_change1
         (fun () ->
           setCountStr (fun (count, str) -> (count + 1, f str)) ;
           None )
-        [|f|] ;
+        f ;
```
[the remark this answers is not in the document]
glennsl: I think that might be the best approach yeah. We should perhaps draw up a a plan for what this compatibility library would look like before we merge this, to make sure we have everything covered.

lib/core.mli
glennsl: It's not that I have a use case in mind, more of a concern that there _ought to be_ a use case for this very weird React API. I just can't understand how that API came about, other than there being some practical reason for it.

lib/core.mli
```diff
@@ -104,254 +104,236 @@ val clone_element : element -> 'props -> element
...
 
-    let use_effect3 callback watchlist =
-      use_effect3_internal Imports.react callback watchlist]
+    let use_effect_on_change ?(before_render = false) f values =
+      (if before_render then _use_layout_effect else _use_effect)
+        Imports.react f values]
 
-val use_effect4 :
-  (unit -> (unit -> unit) option_undefined) -> 'a * 'b * 'c * 'd -> unit
+val use_effect_on_change1 :
```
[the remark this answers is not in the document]
glennsl: Fixed in https://github.com/ml-in-barcelona/jsoo-react/pull/141/commits/cc8db2db68089be6aac837326a5b37e664fee7eb

lib/core.mli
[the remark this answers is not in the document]
glennsl: Fixed in https://github.com/ml-in-barcelona/jsoo-react/pull/141/commits/60214607a2f36b746d22f34579dc774d01a7fc54

glennsl:
No idea what's wrong with the Windows CI today. It seems to fail on everything.

jchavarri:
Thanks!

test/test_ml.ml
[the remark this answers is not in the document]
glennsl: Great idea!

PR #144, opened 2022-02-14
Add global_attributes into both html and svg Props.

davesnx:
Just including all global attributes on svg, so are all available as Props.


lib/dom_global_attributes.ml
```diff
@@ -0,0 +1,29 @@
+include Dom_dsl_core.Prop
+
+let accessKey = string "accessKey"
+
+let className = string "className" (* substitute for "class" *)
+
+let contentEditable = bool "contentEditable"
+
+let contextMenu = string "contextMenu"
```
jchavarri, line 9: This attr is obsolete and will be removed, I wonder if we should just not add it https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contextmenu

lib/dom_global_attributes.ml
```diff
@@ -0,0 +1,29 @@
...
+
+let lang = string "lang"
+
+let role = string "role" (* ARIA role *)
+
+let style = (any "style" : Dom.Style.t -> t)
+
+let spellCheck = bool "spellCheck"
+
+let tabIndex = int "tabIndex"
```
jchavarri, line 27: What is the rule about camelCase / snake_case in general? I assume for DOM related stuff we keep the original names? cc @glennsl

jchavarri (approved):
Looks good! 👍

lib/dom_global_attributes.ml
davesnx: Done in aa8a2e5

lib/dom_html.ml
```diff
@@ -11,36 +12,6 @@ module Prop = struct
 
   let defaultValue = string "defaultValue"
 
-  (* global html attributes *)
```
glennsl, line 14: Hold on, these are global _HTML_ attributes, meaning that they apply to all HTML elements, but not necessarily to any other elements, like SVG. SVG does specify [a set of "core" attributes](https://svgwg.org/svg2-draft/struct.html#CommonAttributes) which overlap quite a bit with these, but neither are a subset or superset of the other.

If some of these are missing from `Svg`, I'd suggest just adding them directly there.

lib/dom_global_attributes.ml
glennsl: Good question. HTML attributes are actually flatcase, so if we were restricted to that we could just ignore the issue altogether. Unfortunately our life is not that simple. SVG actually manages to mix both flatcase, cemaelCase _and_ kebab-case in its attribute naming. And React of course has chosen camelCase even for flatcased names. Since there are other React-specific naming conventions we rely on as well (`className`, `htmlFor` etc.), it might be better to just stick with that, yeah.

Now I wonder if we should have stuck with camelCase for CSS properties too...

lib/dom_html.ml
davesnx: Following the HTML spec, you are totally correct. By the spec SVGs aren't valid with those attributes attached. This isn't the same in React JSX where any of those are available and valid (in their TypeScript definitions as well): https://codesandbox.io/s/mutable-fire-cd5ro?file=/src/App.js

The only attribute that isn't valid on the sandbox is 'accessKey' (wasn't aware of that one).

davesnx: I'm unsure what we prefer, to keep it close to the spec (which seems reasonable for me) or keep it as React. For now, I would argue that all props are available to all elements, which makes HTML validity (or SVG validity) a joke. 

Part of the effort of https://github.com/ml-in-barcelona/jsoo-react/pull/71 and https://github.com/ml-in-barcelona/jsoo-react/pull/66/files was to create a more HTML safe mechanism than reason-react. This effort with the Dom_Html got a little lost.

glennsl: I was thinking it'd be better to just copy the missing one's to `Svg`. The problem with having a set of modules with "global" attributes is that it implies that these apply to everything. And so if at some point in the future we decide to add support for something like VML, for example (for the impending metaverse), or MathML, then odds are these won't all be applicable.

> Part of the effort of https://github.com/ml-in-barcelona/jsoo-react/pull/71 and https://github.com/ml-in-barcelona/jsoo-react/pull/66/files was to create a more HTML safe mechanism than reason-react. This effort with the Dom_Html got a little lost.

Thanks for pointing to these. I've seen this referenced elsewhere, but hadn't actually seen the PRs. They're very well-structured and pleasant to read too! What I still don't understand, however, is the rationale for this. I see correctness mentioned in one of the PRs, but correctness shouldn't be an end in and of itself (although if it's free then I certainly wouldn't mind!) I also don't entirely understand how this PR fits into the picture.

jchavarri: > (for the impending metaverse)

🤣 

> I still don't understand, however, is the rationale for this

I believe (@davesnx can correct me) the rationale was to have a way to help users identify cases where the wrong attribute is being passed to some element.

> but correctness shouldn't be an end in and of itself 

could you elaborate? I don't understand this, I would always assume using correctness to have the tool help users build valid trees would be useful, all other things being equal.

glennsl: Not necessarily. There are many situations where the price of correctness is higher than the benefit. And what is "correct" isn't always well defined. As is the case with HTML, which is an ever-changing living standard. It's also not just that new elements and attributes are added, some also get deprecated. Is it then correct to use a deprecated attribute? Is it correct to use an attribute that has just been standardized but isn't yet supported by any browser?

In addition comes the cost of maintaining such a complex and ever-changing system and the procedures for deciding what to include and what not to include in it. But that's still just one side of the equation, and why I'm asking about the other side. There might be benefits that outweigh these costs, but I can't really say that I see them.

I can see that rejecting attributes that are invalid on certain elements as a use case, but I don't actually see much value in that in practice. Looking at my the codebase I'm working on now, we use very few non-global attributes. There's quite a few `<a href=...>`s and ``s, and some `<input type=...>`s, but that's really about it for HTML. SVG is much more interesting, but the challenge there is to remember the cryptic attribute names in the first place, and what mysterious sequence of characters to put in them, not so much which goes where. And I'll usually create a higher level API so as to avoid having to deal with that directly anyway.

That's not to say I'm against this idea. I just don't have a clear picture of either costs or benefits. I also think some of the costs will depend on what benefits are being targeted. For example, if this would enable completion of the reduced set of attributes that are valid for the current element, that would have significant value to me. But I can't see how to accomplish that without incurring costs that I would consider too high (like changing the DSL to something like `Html.Div.(make Props[| ... |] [])`).

davesnx:
Updated the "missing" props into SVG and leave the discussion of correctness/safety out of this PR.

PR #151, opened 2022-02-25
allow locally abstract type and type constraint on component definitions

glennsl:
This makes the ppx not reject locally abstract type definitions (e.g. `(type a)`) and type constraints on component definitions.

For example, these used to be rejected:

```ml
let%component make (type a) ~(foo : a) : _ = div [||] []
let%component make : type a. foo:a -> _ = fun ~foo:_ -> div [||] []
```

with the error

```
react.component calls can only be on function definitions or component wrappers (forward_ref, memo).
```

The latter also produced the error:

```
Fatal error: exception (Invalid_argument "react.component calls cannot be destructured.")
```

Which has been improved, along with a few other errors, by using `Location.raise_errorf` instead of `invalid_arg`.


davesnx (approved):
Great job!

jchavarri (approved):
This is awesome 🚀

tjltjl:
Really great! I just bumped into this yesterday :D

PR #155, opened 2022-03-06
consistent project-wide formatting

glennsl:
This started with me wanting to pin `ocamlformat` to a specific version because #154 failed CI despite not failing locally because of a new version introducing a different. Then I went on to unify all the `.ocamlformat` files into one project-wide configuration, and finally added some tweaks to it that I've found to work well, because I found the code I added in #154 got absolutely butchered compared to the version I have in my own project.

For comparison:

```ml
let use_effect ~on f =
  let last_value = React.use_ref on in

  (* runs initially *)
  React.use_effect0 (fun () ->
      f ();
      None);

  (* runs always *)
  React.use_effect (fun () ->
      if on <> React.Ref.current last_value then begin
        React.Ref.set_current last_value on;
        f ()
      end;
      None)

```

```ml
let use_effect ~on f =
  let last_value = Core.use_ref on in
  Core.use_effect_once (fun () -> f () ; None) ;
  Core.use_effect_always (fun () ->
      if on <> Core.Ref.current last_value then (
        Core.Ref.set_current last_value on ;
        f () ) ;
      None )
```

The reasoning for the configuration choices is as follows:

- `profile = default` - I find both `default` and `ocamlformat` to be pretty inconsistent, `ocamlformat` more so than `default`, and actually prefer `janestreet` I think, if not for one notable exception: that it doesn't preserve single blank lines in function definitions, which I find absolutely essential for grouping. And there's unfortunately no configuration option for that. Therefore `default` seems like the more sensible baseline.
- `break-separators = before` and `dock-collection-brackets = false` - These put the delimiters and separators of lists, records etc. in front of the items, and in the same column, i.e. Haskell/Elm-style. This makes line-based editing easier while also being more compact, and arguably more readable (once familiar with the style). The downside is that it's quite unfamiliar to people coming from JavaScript and similar languages.
- `exp-grouping = preserve` - Simply enables the use of `begin/end` instead of parentheses, in rare cases where that's more readable.
- `type-decl = sparse` - Always formats variants and records with each item on a separate line instead of trying to fit them on a single line. This is more consistent, easier to read, and makes line-based editing much easier.
- `break-fun-decl = fit-or-vertical` and `break-fun-sig = fit-or-vertical` - Puts function arguments on separate lines if they can't all fit on a single line, instead of trying to fit as much as possible on each line. Easier to read and makes line-based editing easier.

Feel free to disagree though! I find all of these add value individually, and would rather have either one of them than none at all.

Also, this obviously can't be merged unless all other PRs have already been merged, or it would create a huge mess. So Draft for now.

### DIFF
--- a/ppx/html.ml
+++ b/ppx/html.ml
@@ -1,6 +1,13 @@
 let ( & ) = List.append
 
-type attributeType = String | Int | Float | Bool | Style | Ref | InnerHtml
+type attributeType =
+  | String
+  | Int
+  | Float
+  | Bool
+  | Style
+  | Ref
+  | InnerHtml
 
 type eventType =
   | Clipboard
@@ -20,13 +27,25 @@ type eventType =
   | Pointer
   | Drag
 
-type attribute = {type_: attributeType; name: string; jsxName: string}
+type attribute =
+  { type_ : attributeType
+  ; name : string
+  ; jsxName : string
+  }
 
-type event = {type_: eventType; name: string}
+type event =
+  { type_ : eventType
+  ; name : string
+  }
 
-type prop = Attribute of attribute | Event of event
+type prop =
+  | Attribute of attribute
+  | Event of event
 
-type element = {tag: string; attributes: prop list}
+type element =
+  { tag : string
+  ; attributes : prop list
+  }
 
 let attributeReferrerPolicy = String
 (* | Empty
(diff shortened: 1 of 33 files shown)


ppx/html.ml
```diff
@@ -48,415 +67,445 @@ let attributeAnchorTarget = String
...
-  ; Event {name= "onWheelCapture"; type_= Wheel}
-  ; Event {name= "onAnimationStart"; type_= Animation}
-  ; Event {name= "onAnimationStartCapture"; type_= Animation}
-  ; Event {name= "onAnimationEnd"; type_= Animation}
-  ; Event {name= "onAnimationEndCapture"; type_= Animation}
-  ; Event {name= "onAnimationIteration"; type_= Animation}
-  ; Event {name= "onAnimationIterationCapture"; type_= Animation}
-  ; Event {name= "onTransitionEnd"; type_= Transition}
-  ; Event {name= "onTransitionEndCapture"; type_= Transition} ]
+  [ Event { name = "onCopy"; type_ = Clipboard }
```
davesnx, line 70: That case is really weird. ocamlformat and the highlight don't like each other.

davesnx (approved):
Great effort @glennsl

jchavarri (approved):
I guess this would be nice to have to fix ci issues in `main`.

glennsl:
Rebased on main

jchavarri:
Thanks!

PR #156, opened 2022-03-13
dsl: fix react runtime warning when using maybe with None

### LINKED ISSUE #130: React shows warning when using `maybe` + None
When using a prop with `maybe` and passing `None` (e.g. `maybe className None`), React.js will emit a warning:

```
Warning: Invalid attribute name: ``
    in div
```

This happens in ml syntax:

https://github.com/ml-in-barcelona/jsoo-react/blob/d3924adb125a8a7fadb8c2535486127c9b05c1fa/test/test_ml.ml#L704

And Reason syntax:

https://github.com/ml-in-barcelona/jsoo-react/blob/d3924adb125a8a7fadb8c2535486127c9b05c1fa/test/test_reason.re#L1156

I tried using any other string like `_` or `_unused` and in those cases React does not complain 🤔 Another alternative would be to conditionally push to the array.

glennsl:
Fixes #130

### DIFF
--- a/lib/dom_dsl_core.ml
+++ b/lib/dom_dsl_core.ml
@@ -10,7 +10,10 @@ module Prop = struct
   let event key (f : _ Event.synthetic -> unit) =
     any key (Js_of_ocaml.Js.wrap_callback f)
 
-  let maybe prop = function Some value -> prop value | None -> any "" ""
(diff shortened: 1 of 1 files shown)


jchavarri (approved):
Thanks!

PR #157, opened 2022-03-13
dsl: add classNames helper to Html and Svg props

glennsl:
I've found it very useful to be able to pass a list of class names, rather than having to build a class string manually. This is the simplest version of that, but I could imagine something more sophisticated to support optional class names for example.

Currently, to pass an optional class name you'd have to do something like:

```ml
div [| classNames [ "widget"; if disabled then "disabled" else "" ] |] [ ... ]
```

but could perhaps instead be:

```ml
div [| classNames [ "widget"; class_if disabled "disabled" ] |] [ ... ]
```

But this could be added in a way that maintains backwards-compatibility, so not all that important to do now I think.

### DIFF
--- a/lib/dom_html.ml
+++ b/lib/dom_html.ml
@@ -16,6 +16,7 @@ module Prop = struct
   let accessKey = string "accessKey"
   let className = string "className" (* substitute for "class" *)
 
+  let classNames names = className (String.concat " " names)
   let contentEditable = bool "contentEditable"
   let contextMenu = string "contextMenu"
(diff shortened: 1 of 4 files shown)


glennsl:
> Do you think this should be part of jsoo-react? Or rather a 3rd party lib like [rescript-classnames](https://github.com/shakacode/rescript-classnames)
?

I'm thinking it might as well be, since I imagine pretty much everyone would want this. Depends a bit on how opinionated the solution is though. Could also do a halfway solutions by putting it in a separate namespace, temporarily or permanently. but either way there's no rush I think.

PR #158, opened 2022-03-13
dom: add create_portal

glennsl:
For easy reference: https://reactjs.org/docs/portals.html

rescript-react binding: https://github.com/rescript-lang/rescript-react/blob/6156e3b0f8fd23f7eb309e0e8de0d42612a51180/src/ReactDOM.res#L30-L31


davesnx (approved):

jchavarri (commented):
❤️

PR #160, opened 2022-03-19
dsl: expose ARIA attributes

glennsl:
Includes the ARIA attributes that were added in #145 in the `Dom_html` and `Dom_svg` module interfaces, so that they can actually be used! :wink:


davesnx:
My bad, I didn't try myself after the PR with a real project. Thanks for fixing it!

jchavarri:
Thanks!

PR #164, opened 2022-04-03
dom: add missing svg props

glennsl:
I might find more as I continue to dig into SVG in the near future, so keeping this a draft for now.


jchavarri:
Thank you.

PR #165, opened 2022-04-03
A few ergonomic additions for dealing with conditional rendering

glennsl:
The addition of `maybe` for elements is a slightly breaking change, as it shadows `maybe` for `Prop`. It's a very low effort fix though.

Another ergonomic addition I've considered is making the attributes optional with an `~a` labeled argument, as in tyxml, and the addition of short-hand labeled arguments for `id` and `className` attributes, as I've noticed that I very rarely use any attributes other than `className`, which is used quite often. But I'm thinking it might be better to keep the DSL simple and add a ppx layer for even better ergonomics instead without any runtime cost.

### DIFF
--- a/lib/dom_dsl_core.ml
+++ b/lib/dom_dsl_core.ml
@@ -36,4 +36,5 @@ module Common = struct
   let string = Core.string
   let int = Core.int
   let float = Core.float
+  let maybe f = function Some value -> f value | None -> none
 end
--- a/lib/dom_dsl_core.ml
+++ b/lib/dom_dsl_core.ml
@@ -32,6 +32,7 @@ module Common = struct
   end
 
   let fragment children = Core.Fragment.make ~children ()
+  let none = fragment []
   let string = Core.string
   let int = Core.int
   let float = Core.float
--- a/ppx/ppx.ml
+++ b/ppx/ppx.ml
@@ -943,9 +943,7 @@ let jsxMapper () =
         match str_label with
         | Str_label.Optional str ->
             [%expr
-              maybe
-                [%e Exp.ident { loc; txt = Ldot (Lident "Prop", str) }]
-                [%e value]]
+              Prop.(maybe [%e Exp.ident { loc; txt = Lident str }]) [%e value]]
         | Labelled str ->
             [%expr
               [%e Exp.ident { loc; txt = Ldot (Lident "Prop", str) }] [%e value]]
--- a/lib/dom_dsl_core.mli
+++ b/lib/dom_dsl_core.mli
@@ -25,6 +25,7 @@ module Common : sig
   end
 
   val fragment : Core.element list -> Core.element
+  val none : Core.element
   val string : string -> Core.element
   val int : int -> Core.element
   val float : float -> Core.element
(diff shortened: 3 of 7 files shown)


glennsl:
I just realized that `React.null` exists, and should probably be copied to the DSL. `null` seems like a pretty bad name though, unless you want to optimize for familiarity with JavaScript's `null`. So not sure if it should be 'none' or 'null' in the DSL.

lib/dom_dsl_core.ml
```diff
@@ -32,7 +32,9 @@ module Common = struct
   end
 
   let fragment children = Core.Fragment.make ~children ()
+  let none = fragment []
```
jchavarri, line 35: Why not
```suggestion
  let none = Core.null
```
?

As you mentioned in your comment.

ppx/ppx.ml
```diff
@@ -943,9 +943,7 @@ let jsxMapper () =
         match str_label with
         | Str_label.Optional str ->
             [%expr
-              maybe
-                [%e Exp.ident { loc; txt = Ldot (Lident "Prop", str) }]
-                [%e value]]
+              Prop.(maybe [%e Exp.ident { loc; txt = Lident str }]) [%e value]]
```
jchavarri, line 946: The previous implementation was more conservative right? Could not there be shadowing issues down the line caused by opening of `Prop` fully?

I would rather keep the ppx super accurate with namespacing, avoiding usages of `open` as much as possible:

```suggestion
             Prop.maybe
                [%e Exp.ident { loc; txt = Ldot (Lident "Prop", str) }]
                [%e value]]
```

lib/dom_dsl_core.ml
```diff
@@ -32,7 +32,9 @@ module Common = struct
...
   let string = Core.string
   let int = Core.int
   let float = Core.float
+  let maybe f = function Some value -> f value | None -> none
```
jchavarri, line 39: I am not sure this helper is something that should be part of the DSL library scope 🤔 It seems a function that can be easily added to userland. Or maybe we can start a new library for this kind of helpers?

jchavarri (commented):
Thanks for this! A few thoughts.

lib/dom_dsl_core.ml
glennsl: Just because I didn't remember it existed when I wrote this, and now that I do I wonder if it should be called `none` or `null` in the DSL. I'm leaning towards `none` though.

ppx/ppx.ml
glennsl: These seem exactly equivalent to me. `Prop.maybe Prop.foo value`, `Prop.(maybe foo) value` and `(let open Prop in maybe foo) value` all do the exact same thing. But if you think this change makes it clearer I'd be happy to accept it.

lib/dom_dsl_core.ml
glennsl: It could, but then you have another module you'd have to open. It's also a bit inconsistent to provide this functionality for props, but not elements.

I don't feel very strongly about it, but I do think it'd be better to have this out of the box, assuming the design isn't too controversial and prone to change.

Issue #79, opened 2021-12-07
Set displayName in dev mode


jchavarri:
It'd be nice to call `displayName` when using `profile=dev` mode in Dune, so that we can see nice component names in browser when using React extension.

Example: https://github.com/utkarshkukreti/reaml/blob/5640a36293ba2a765171225deddb644f4d6c5d26/src/Reaml_Core.ml#L248

davesnx:
Plus one to this, using the React extension it might be a good idea to visualize the React Tree. Not useful to see the props, but at least is something!

PR #166, opened 2022-04-14
ppx: set display name on let%component

glennsl:
Not sure how to test it, but this produces react stack traces such as this:

```
The above error occurred in the <Dune__exe__HelloWorldOCaml.Foo.make.(fun)> component:
    in Dune__exe__HelloWorldOCaml.Foo.make.(fun)
    in div (created by Dune__exe__Main.make.(fun))
    in div (created by Dune__exe__Main.make.(fun))
    in div (created by Dune__exe__Main.make.(fun))
    in div (created by Dune__exe__Main.make.(fun))
    in Dune__exe__Main.make.(fun)
```

Also note that the lower bound on OCaml was bumped to 4.12 to be able to use `__FUNCTION__`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,4 +49,4 @@ $(opam_file): dune-project ## Update the package dependencies when new deps are 
 
 init: ## Create a local opam switch and setups githooks
 	git config core.hooksPath .githooks
-	opam switch create . 4.10.0 --deps-only --with-test
+	opam switch create . --deps-only --with-test
--- a/ppx/test/pp_ocaml.expected
+++ b/ppx/test/pp_ocaml.expected
@@ -41,7 +41,9 @@ let make =
                  (Props : < .. >  Js_of_ocaml.Js.t) (fun x -> x#name)) in
     fun ?name ->
       fun ?key ->
-        fun () -> React.create_element make (make_props ?key ?name ()))
+        fun () ->
+          React.set_display_name make __FUNCTION__;
+          React.create_element make (make_props ?key ?name ()))
     [@merlin.hide ])
 let make =
   let make_props
@@ -87,7 +89,9 @@ let make =
                      (fun x -> x#children)) in
     fun ~children ->
       fun ?key ->
-        fun () -> React.create_element make (make_props ?key ~children ()))
+        fun () ->
+          React.set_display_name make __FUNCTION__;
+          React.create_element make (make_props ?key ~children ()))
     [@merlin.hide ])
 let make =
   let make_props
@@ -131,7 +135,9 @@ let make =
                      (fun x -> x#children)) in
     fun ~children ->
       fun ?key ->
-        fun () -> React.create_element make (make_props ?key ~children ()))
+        fun () ->
+          React.set_display_name make __FUNCTION__;
+          React.create_element make (make_props ?key ~children ()))
     [@merlin.hide ])
 let make =
   let make_props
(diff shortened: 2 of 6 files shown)


glennsl:
CI errors look like flukes to me :shrug:

glennsl:
MacOS failure was indeed a fluke. Windows failrue seems to be an issue with `setup-ocaml`: https://github.com/ocaml/setup-ocaml/issues/479

Makefile
```diff
@@ -49,4 +49,4 @@ $(opam_file): dune-project ## Update the package dependencies when new deps are
 
 init: ## Create a local opam switch and setups githooks
 	git config core.hooksPath .githooks
-	opam switch create . 4.10.0 --deps-only --with-test
+	opam switch create . --deps-only --with-test
```
jchavarri, line 52: Why not updating version here?

ppx/test/pp_ocaml.expected
```diff
@@ -41,7 +41,9 @@ let make =
                  (Props : < .. >  Js_of_ocaml.Js.t) (fun x -> x#name)) in
     fun ?name ->
       fun ?key ->
-        fun () -> React.create_element make (make_props ?key ?name ()))
+        fun () ->
+          React.set_display_name make __FUNCTION__;
```
jchavarri, line 45: I wonder if there is a way to call `set_display_name` from outside this function, to avoid wasted work. Or is it necessary to do so every time a new element is created?

jchavarri (approved):
Thanks for this!

I don't see any blockers to merge this, just have a question,

Makefile
glennsl: It just seems unnecessary. I also think not specifying it as an invariant might make upgrading easier, since then you shouldn't need to reinstall the whole switch. But I haven't actually tried that.

ppx/test/pp_ocaml.expected
glennsl: It most likely could, but I'm struggling to figure out where that would be in the ppx code. We could probably also mitigate the impact by only emitting this in dev mode, but I'm not sure how to do that either. Yet.

Makefile
jchavarri: I was thinking if there is no version explicitly mentioned in `make init` command, one would install 4.14 (or whatever is latest) just to switch back to 4.12 or whatever we are using for development 🤔

jchavarri: Actually, on second thought, what would happen is that we'd develop with latest version of the compiler, while ci is tied to 4.12. Maybe we should leave it like that (to potentially detect issues in development asap), and enable multi-version ci in the future.

In any case, none of this is blocker for this PR.

jchavarri:
@glennsl this PR would close https://github.com/ml-in-barcelona/jsoo-react/issues/79, right?

Makefile
glennsl: It actually installed 4.13.1 for me, for some reason. We could probably add a constraint on `dev` or `with-test` to have more precise control over this as well.

glennsl:
> @glennsl this PR would close #79, right?

Seems so! Need to figure out how to detect if we're in dev mode though.

ppx/test/pp_ocaml.expected
glennsl: Fixed in https://github.com/ml-in-barcelona/jsoo-react/pull/166/commits/d0cc8c38d163be2ec2ae86279c03d95ad93753eb

glennsl:
Now it'll only set the display name if the environment variable `JSOO_REACT_DEV` set.

jchavarri:
Thanks!

PR #178, opened 2023-03-27
Prefix Option and Array with Stdlib

davesnx:
- Prefix Stdlib modules with Stdlib so core/container users won't shadow
- Update dune-project to match with recent opam updates (https://github.com/ml-in-barcelona/jsoo-react/pull/177). Maybe we could add a check for this in CI?
- I had the switch very outdated and had a hard time pushing broken formatting, I did the following to avoid this issue
  - Specified 4.12 in Makefile
  - Specified Reason version 3.8.2 (with test)
  - Specified ocamlformat version 0.21.0 (with test)

### DIFF
--- a/ppx/test/pp_ocaml.expected
+++ b/ppx/test/pp_ocaml.expected
@@ -16,11 +16,8 @@ let make =
               [|("key",
                   (inject
                      (Js_of_ocaml.Js.Optdef.option
-                        (Option.map Js_of_ocaml.Js.string key))));("name",
-                                                                    (
-                                                                    inject
-                                                                    name))|]
-    [@@merlin.hide ] in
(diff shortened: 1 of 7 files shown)


ppx/test/pp_ocaml.expected
```diff
@@ -16,11 +16,8 @@ let make =
               [|("key",
                   (inject
                      (Js_of_ocaml.Js.Optdef.option
-                        (Option.map Js_of_ocaml.Js.string key))));("name",
```
jchavarri, line 19: Did anything change in the tests? Hard to tell with the refmt changes.

jchavarri (approved):

ppx/test/pp_ocaml.expected
davesnx: Just the Module added

PR #180, opened 2023-04-07
Update README.md

davesnx:
Added installation from opam https://github.com/ocaml/opam-repository/pull/23646
Updated the template to ensure everything works as expected: https://github.com/ml-in-barcelona/jsoo-react-template/commit/603e1fb66eee07a24d8600838c9cc1624c0c9724


jchavarri:
We should update usage instructions too, as pinning won't be needed any more:

https://github.com/ml-in-barcelona/jsoo-react/blob/57f42ebf664235e01af136a789a24e57d05f884f/README.md#L24

PR #185, opened 2023-10-09
Remove upper bounds for OCaml 5

davesnx:
- Update OCaml version
- Remove other upper bounds

This was needed since I want to use OCaml 5 for query-json here: https://github.com/davesnx/query-json/pull/38


michael-schwarz:
Are there plans for merging this? It looks innocuous enough?

davesnx:
You are right @michael-schwarz

michael-schwarz:
Thanks!

davesnx:
You would need to pin from #main to use with this changes!

